PR #10, opened 2015-11-10
Support smarter logging / Request timeouts

sethrosetter:
- If a test fails, the framework will snapshot the logs from sync_gateways from that test to /tmp
- Add 30 sec timeout on user requests

### DIFF
--- a/functional_tests/cluster_setup.py
+++ b/functional_tests/cluster_setup.py
@@ -1,11 +1,25 @@
 import pytest
 
 from lib.cluster import Cluster
-
+from provision.fetch_sg_logs import fetch_sync_gateway_logs
 
 @pytest.fixture()
-def cluster():
+def cluster(request):
+
+    def fetch_logs():
+
+        # Fetch logs if a test fails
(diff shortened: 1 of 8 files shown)


functional_tests/cluster_setup.py
```diff
@@ -1,11 +1,25 @@
...
+        # Fetch logs if a test fails
+        if request.node.rep_call.failed:
+            # example nodeid: tests/test_single_user_multiple_channels.py::test_1
+            remove_slash = request.node.nodeid.replace("/", "-")
+            test_id_elements = remove_slash.split("::")
+            log_zip_prefix = "{0}-{1}".format(test_id_elements[0], test_id_elements[1])
+            fetch_sync_gateway_logs(log_zip_prefix)
+
+    request.addfinalizer(fetch_logs)
+
```
ashvindersingh: Is there an option to disable collecting logs at the end of tests. For example, while developing tests, need to run script multiple times and may not need to collect any logs on failure.

sethrosetter:
Reopen at a later date

PR #21, opened 2015-11-12
Add cache timeout test

sethrosetter:
- Add 1k users 5k docs channel cache overload test
- Simplify server provisioning
- Improved test output for changes request sanity checks
- Minor ansible tweaks to sg installation
- _changes request adds ability for filters
- docs id verifier for changes feed

### DIFF
--- a/lib/user.py
+++ b/lib/user.py
@@ -8,9 +8,11 @@ import time
 import base64
 import uuid
 import re
+from lib.document import Document
 from collections import defaultdict
 from responseprinter import ResponsePrinter
 
+PRINTER = ResponsePrinter()
 
 class User:
     def __init__(self, target, db, name, password, channels):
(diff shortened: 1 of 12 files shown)


lib/user.py
```diff
@@ -8,9 +8,11 @@
 import base64
 import uuid
 import re
+from lib.document import Document
 from collections import defaultdict
 from responseprinter import ResponsePrinter
 
+PRINTER = ResponsePrinter()
```
tleyden: What about `response_printer` instead of `PRINTER`?

sethrosetter: I was just emphasizing that it is global and will be shared among threads but if that is the convention for globals, I'm in agreement.

tleyden: I don't see anything about using all caps here:  https://www.python.org/dev/peps/pep-0008/#id35

sethrosetter: @tleyden fixed. Also changed the name as it is more than a "response" printer now

tleyden:
LGTM -- go ahead and merge from my POV (it's assigned to @ashvindersingh so I'm holding off on merging)

sethrosetter:
@tleyden Thanks. I will was just looking for a quick once over in ansible land. But I appreciate the overall feedback!

PR #22, opened 2015-11-12
Fixes #20 - Index writer SG config

### LINKED ISSUE #20: Ability to specify index writers (mix of index-writers and non-writers)
In order to be able to gather performance data as in https://github.com/couchbase/sync_gateway/issues/1272, the perfcluster work needs to be able to support "mixed mode" where:

* Certain instances are tagged as index writers, and other instances are non-writers.  
* The gateload/gatling load generators should only be pointed at the non-writers 
* The index writers will need to have their JSON config changed to set indexwriter = true (the rest of the config would be identical)

Steps that need to happen:

1. Modify sg configs and restart sgs
1. Generate gateload configs to point only at non writers

tleyden:
https://github.com/couchbaselabs/sync-gateway-testcluster/issues/20

### DIFF
--- a/provision/ansible/playbooks/tasks/start-sync-gateway.yml
+++ b/provision/ansible/playbooks/tasks/start-sync-gateway.yml
@@ -1,7 +1,4 @@
 - debug: msg="Starting sync_gateway with config {{ sync_gateway_config_filepath }}"
-- name: copy sync gateway config to host
-  template: src={{ sync_gateway_config_filepath }} dest=/home/sync_gateway/sync_gateway.json owner=sync_gateway group=sync_gateway mode=0644 force=true
-
   # install service and launch
 - name: install sync gateway service
   shell: ./sync_gateway_service_install.sh --sgpath=/home/sync_gateway/sync_gateway chdir=/home/centos/sync_gateway/service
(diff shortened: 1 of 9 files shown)


provision/ansible/playbooks/tasks/start-sync-gateway.yml
```diff
@@ -1,6 +1,3 @@
-# Add sync_gateway user, copy new config
-- name: add sync gateway user
-  user: name=sync_gateway createhome=yes
 - name: make service install script executable
```
sethrosetter: Could we also pull these steps out into the install-source and install-package script as well?

provision/ansible/playbooks/tasks/start-sync-gateway.yml
```diff
@@ -8,9 +5,6 @@
 - name: change permissions of sync gateway executable
   file: path=/home/sync_gateway/sync_gateway mode=a+x
 - debug: msg="Starting sync_gateway with config {{ sync_gateway_config_filepath }}"
```
sethrosetter: Could we also pull these steps out into the install-source and install-package script as well?

sethrosetter:
@tleyden please rebase on master. Thanks!

tleyden:
@sethrosetter ok this is rebased!  Can you implement the change in your suggestion and push up a new commit?

PR #28, opened 2015-11-13
Remove ini

sethrosetter:
- 'provisioning_config' is now the source of truth in 
   - provisioning 
   - funtional test framework
   - gatling / gateload config generation


sethrosetter:
@ashvindersingh please take a look at this as well

lib/syncgateway.py
```diff
@@ -23,3 +23,6 @@ def start(self, config):
 
     def restart(self, config):
         orchestration.syncgatewayactions.restart(self.host_name, config)
+
+    def __str__(self):
```
ashvindersingh: If this implemented for debugging purposes then I would recommend using __repr__

lib/cluster.py
```diff
@@ -1,20 +1,66 @@
...
+            sys.exit(1)
+
+        i = ansible.inventory.Inventory(host_list=hostfile)
+        hosts = i.get_group(tag).get_hosts()
+        return [host.get_variables() for host in hosts]
 
     def reset(self, config):
         clusteractions.reset(config)
 
+    def __str__(self):
```
ashvindersingh: __repr__ will give more accurate description of the object

sethrosetter:
@ashvindersingh updated per your request

ashvindersingh:
Thanks!

PR #38, opened 2015-11-16
Fixes #37 (Tests Hang) by adding default timeout

### LINKED ISSUE #37: Tests hang
If the sync gateway changes feed hangs indefinitely, the `test_single_user_single_channel.py` test will also hang indefinitely (there might be some eventual timeout, but it's a long time)

tleyden:
Conflicts:
	lib/admin.py
	lib/user.py

### DIFF
--- a/lib/admin.py
+++ b/lib/admin.py
@@ -4,6 +4,7 @@ import concurrent.futures
 
 from lib.user import User
 from lib.scenarioprinter import ScenarioPrinter
+from lib import settings
 
 # Admin
 # GET /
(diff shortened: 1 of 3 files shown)


tleyden:
@sethrosetter I hadn't pulled, and didn't see you'd already implemented this.  I implemented it slightly different and added a settings module so that the same constant can be used everywhere.  Per [pep-8](https://www.python.org/dev/peps/pep-0008/#constants), named it all-caps w/ underscores.

PR #42, opened 2015-11-17
Added Retry logic for http requests

ashvindersingh:
Also, updated one test scenario with parameterization.


sethrosetter:
Can you rebase with changes from master and use the settings.HTTP_REQ_TIMEOUT for the timeout line?
ex. https://github.com/couchbaselabs/sync-gateway-testcluster/blob/master/lib/user.py#L43

PR #43, opened 2015-11-17
Fix lib.admin users

sethrosetter:
- admin.register_user now returns a User object
- admin.register_bulk_users now returns a list of User objects
- added settings.MAX_REQUEST_WORKERS = 100 to tune concurrent requests across library


ashvindersingh:
looks good

PR #48, opened 2015-11-18
http request retry mechanism

ashvindersingh:
This PR implements: 
- HTTP requests retry logic
- Parameterization of single_user_single_channel_doc_updates

### DIFF
--- a/functional_tests/test_single_user_single_channel_doc_updates.py
+++ b/functional_tests/test_single_user_single_channel_doc_updates.py
@@ -1,23 +1,35 @@
 import time
+<<<<<<< HEAD
+=======
 import pytest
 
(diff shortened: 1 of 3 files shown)


tleyden:
I think this PR needs more cleanup, if you just squash those two commits into a single commit and do a force push, that should do it.

sethrosetter:
Should we implement this for all the requests or no?

ashvindersingh:
@sethrosetter By default the retry mechanism is for both adding and updating docs.

sethrosetter:
@ashvindersingh Do we want retry on bulk_docs or no?

ashvindersingh:
@sethrosetter I think we should do

PR #51, opened 2015-11-20
Unify verification

sethrosetter:
!! This PR fundamentally changes how we are validating changes feeds. Please review the verify_changes function and the user class throughly. These changes will allow unified testing across all the scenarios and opens up the possibility for simple concurrent changes processing. 

### Test Scenarios
- ported to use new validation
- ground work for further scenario parametrization

### Framework enhancements
- testing doc ids, latest revs, and number of update all from a unified location
- fixed issue where bulk_docs was storing possibly incorrect values in user cache
- user.add_doc(num) adds uuid name by default unless you specify a user.add_doc(num, name_prefix="text-1")
- remove user.get_num_docs(), not used anywhere
- hopefully resolve intermittent ssh changes


adamcfraser:
I don't follow why the new verify_changes takes a list of users, but only has a single expected doc set/count to compare against.  If it doesn't handle multiple users with different channel sets, is there a reason to support multiple users requesting the same data?

sethrosetter:
@adamcfraser I was thinking at level of verifying common subsets of users i.e the list would be users subscribed to the same channel or same role. I see what you are saying though. I will talk with @ashvindersingh about this and think about this a bit more.

lib/user.py
```diff
@@ -55,34 +58,45 @@ def add_bulk_docs(self, doc_ids):
...
 
-        r = session.post("{0}/{1}/_bulk_docs".format(self.target.url, self.db), headers=self._headers, data=data, timeout=settings.HTTP_REQ_TIMEOUT)
-        scenario_printer.print_status(r)
-        r.raise_for_status()
+        resp = session.post("{0}/{1}/_bulk_docs".format(self.target.url, self.db), headers=self._headers, data=data, timeout=settings.HTTP_REQ_TIMEOUT)
+        scenario_printer.print_status(resp)
+        resp.raise_for_status()
+        resp_json = resp.json()
+
+        if len(resp_json) != len(doc_ids):
```
sethrosetter: raise exception here

lib/user.py
```diff
@@ -55,34 +58,45 @@ def add_bulk_docs(self, doc_ids):
...
+            print("Number of bulk docs inconsistent: resp_json['docs']:{} doc_ids:{}".format(len(resp_json), len(doc_ids)))
 
-        if r.status_code == 201:
-            for doc_id in doc_ids:
-                self.cache[doc_id] = 0
+        # Store docs from response in user's cache and save list of ids to return
+        bulk_docs_ids = []
+        if resp.status_code == 201:
+            for doc in resp_json:
+                self.cache[doc["id"]] = doc["rev"]
```
ashvindersingh: save the rev-id prefix instead of full revision

lib/user.py
```diff
@@ -124,13 +140,18 @@ def update_doc(self, doc_id, num_revision=1):
                 put_resp = session.put(doc_url, headers=self._headers, data=body, timeout=settings.HTTP_REQ_TIMEOUT)
 
                 if put_resp.status_code == 201:
+
                     data = put_resp.json()
-                    if data["rev"]:
-                        self.cache[doc_id] = data["rev"]  # init doc revisions to 0
-                    else:
-                        print "Error: Did not fine _rev after Update response"
+                    assert "rev" in data.keys()
```
ashvindersingh: We can use exception here instead assert

lib/user.py
```diff
@@ -124,13 +140,18 @@ def update_doc(self, doc_id, num_revision=1):
...
+
+                    # Update revision number for stored doc id
+                    self.cache[doc_id] = data["rev"]
+
                 put_resp.raise_for_status()
             resp.raise_for_status()
+
+        # Return doc_id, doc_rev kvp
+        return self.cache[doc_id]
```
ashvindersingh: Return list of newly added doc-ids/rev key value pair.

lib/verify.py
```diff
@@ -0,0 +1,63 @@
+
+def verify_change_with_filter(doc_name_pattern):
+    pass
+
+def verify_changes(users, expected_num_docs, expected_num_updates, expected_docs):
+
+    if type(users) is not list:
```
ashvindersingh: Coalesce single user into a list

lib/verify.py
```diff
@@ -0,0 +1,63 @@
+
+def verify_change_with_filter(doc_name_pattern):
+    pass
+
+def verify_changes(users, expected_num_docs, expected_num_updates, expected_docs):
```
ashvindersingh: rename expected_num_updates to expected_num_revisions

lib/verify.py
```diff
@@ -0,0 +1,63 @@
...
+        # Assert there are no duplicates in expected doc ids
+        assert len(expected_doc_ids) == len(set(expected_doc_ids))
+
+        # Get ids from all changes results
+        changes_doc_ids = list()
+        for result in changes_results:
+            changes_doc_ids.append(result["id"])
+
+        # Assert there are no duplicates in changes doc ids
+        assert len(changes_doc_ids) == len(set(changes_doc_ids))
```
ashvindersingh: We can have list of errors and then check later instead of asserts

lib/verify.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Assert the expected doc ids and changes doc ids are the same
+        assert set(expected_doc_ids) == set(changes_doc_ids)
+
+        for result in changes_results:
+            # Compare revision number for id
+            assert expected_docs[result["id"]] == result["rev"]
+
+            # Check number of expected updates matched the updates on the _changes doc
+            assert expected_num_updates == result["updates"]
```
ashvindersingh: expected_num_revisions == changed.feed(_rev) prefix.

PR #59, opened 2015-11-30
Create cbgt-bucket at startup

tleyden:
It seems a bit strange to create 2 buckets and then have them later re-created with 3 buckets.


sethrosetter:
@tleyden Can you update here as well https://github.com/couchbaselabs/sync-gateway-testcluster/blob/master/provision/ansible/playbooks/install-sync-gateway-package.yml#L30

tleyden:
Here's the problem it solves: https://github.com/couchbaselabs/sync-gateway-testcluster/issues/60

tleyden:
@sethrosetter done

sethrosetter:
looks good to me

PR #67, opened 2015-12-02
Sync sanity scenarios

sethrosetter:
- Add sanity test for access() in sync function
- Add sanity test for channel() in sync function
- Add sanity test for role() in sync function
- Add requireRoles() sanity test
- Add ignore_rev_ids=False to verify changes. This is to ignore rev_id checking for the situation where the user verify their changes includes doc that they did not create or update
- Add verify_docs_removed function (Needs REVIEW) to check a users changes feed that had documents in channels that were changed

### DIFF
--- a/lib/verify.py
+++ b/lib/verify.py
@@ -22,7 +22,106 @@ def verify_same_docs(expected_num_docs, doc_dict_one, doc_dict_two):
     print(" -> doc_dict_one == doc_dict_two expected (num_docs: {})".format(expected_num_docs))
 
 
-def verify_changes(users, expected_num_docs, expected_num_revisions, expected_docs):
+def verify_docs_removed(users, expected_num_docs, expected_docs):
+
+    # Verifies that the expected_docs have all been flagged with _removed = true
+    # Also verifies no duplication of changes results and set equality of the expected doc_ids
+    # and the ids returned from the _changes feed
+
+    errors = {
+        "unexpected_changes_length": 0,
+        "invalid_expected_docs_length": 0,
+        "duplicate_expected_ids": 0,
+        "duplicate_changes_doc_ids": 0,
+        "expected_doc_ids_differ_from_changes_doc_ids": 0,
+        "doc_not_removed": 0,
+        "invalid_rev_id": 0
+    }
+
+    if type(users) is list:
+        user_list = users
+    else:
+        # Allow a single user to be passed
+        user_list = list()
+        user_list.append(users)
+
+    if type(expected_docs) is not dict:
+        raise Exception("Make sure 'expected_docs' is a dictionary")
+
+    for user in user_list:
+
+        # get changes feed
+        changes = user.get_changes(include_docs=True)
+        results = changes["results"]
+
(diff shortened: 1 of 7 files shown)


sethrosetter:
@ashvindersingh this has been rebased off of master

lib/verify.py
```diff
@@ -22,7 +22,106 @@ def verify_same_docs(expected_num_docs, doc_dict_one, doc_dict_two):
...
+
+        # Get ids from expected docs
+        expected_doc_ids = expected_docs.keys()
+
+        # Assert there are no duplicates in expected doc ids
+        if len(expected_doc_ids) != len(set(expected_doc_ids)):
+            errors["duplicate_expected_ids"] += 1
+
+        # Get ids from all changes results
+        changes_doc_ids = [result["id"] for result in changes_results]
```
ashvindersingh: this can be done when parsing the changes feed results.

lib/verify.py
```diff
@@ -22,7 +22,106 @@ def verify_same_docs(expected_num_docs, doc_dict_one, doc_dict_two):
...
+            elif result["removed"] is True:
+                num_doc_removed += 1
+
+            # Compare revision number for id
+            if expected_docs[result["id"]] != result["rev"]:
+                errors["invalid_rev_id"] += 1
+
+            # TODO - maybe try to ping doc endpoint and asser 4XX response?
+
+        print(" -> REMOVED |{0}| expected (num_docs: {1}) _changes (num_docs: {2}, num_removed: {3})".format(
```
ashvindersingh: we should use log function instead of print.

lib/verify.py
```diff
@@ -22,7 +22,106 @@ def verify_same_docs(expected_num_docs, doc_dict_one, doc_dict_two):
     print(" -> doc_dict_one == doc_dict_two expected (num_docs: {})".format(expected_num_docs))
 
 
-def verify_changes(users, expected_num_docs, expected_num_revisions, expected_docs):
+def verify_docs_removed(users, expected_num_docs, expected_docs):
```
ashvindersingh: this function should verify that tombstone revision doesn't change after the doc has been marked deleted.

lib/verify.py
```diff
@@ -90,10 +189,15 @@ def verify_changes(users, expected_num_docs, expected_num_revisions, expected_do
         if set(expected_doc_ids) != set(changes_doc_ids):
             errors["expected_doc_ids_differ_from_changes_doc_ids"] += 1
 
+        if ignore_rev_ids:
+            print("WARNING: Ignoring rev id verification!!")
```
ashvindersingh: same as above comment, use log instead of print.

lib/verify.py
```diff
@@ -48,6 +151,7 @@ def verify_changes(users, expected_num_docs, expected_num_revisions, expected_do
         user_list.append(users)
 
     if type(expected_docs) is not dict:
+        log.error("expected_docs is not a dictionary")
```
ashvindersingh: It would be helpful to also log the user-id

PR #71, opened 2015-12-03
Wait for alternative

sethrosetter:


tleyden:
That should work, just curious .. why `lsof` instead of curl?

Can you squash down to 1 commit?

sethrosetter:
@tleyden no real reason for doing so. Squashed

PR #73, opened 2015-12-04
Add smarter logging.

sethrosetter:
- moved fetch_logs to utilities folder
- enable log capture by default, can disable in functional_tests/setttings.py
- run_tests.py that will clear logs before run to prevent Jenkins and local logging bloat


ashvindersingh:
looks good!

PR #103, opened 2015-12-10
Ash bucket online offline 1

ashvindersingh:
- Added webhook support


tleyden:
@ashvindersingh is there a typo in the title?

sethrosetter:
@ashvindersingh is 'test_framework_ip' the web hook server? I think we should rename it so it is less ambiguous in the provisioning scripts.

PR #181, opened 2015-12-29
Add missing REST endpoints for db online / offline tests

### LINKED ISSUE #178: [DB Online / Offline] Make sure to hit missing critical REST endpoints
/{db}/{local-doc-id}
/{db}/_bulk_get

sethrosetter:
Fixes #178 

Framework enhancements 
- Add user.get_docs(doc_ids), which used _bulk_get in the implementation
- Add user.add_doc, which uses a POST to /{db}/ and lets sync_gateway generate the doc id
- Support for local docs in add_doc("_local/{name}")
- Add user.get_doc()

Test enhancement
- Hit these endpoints in db online / offline scenarios

### DIFF
--- a/functional_tests/test_db_online_offline.py
+++ b/functional_tests/test_db_online_offline.py
@@ -18,7 +18,7 @@ from multiprocessing.pool import ThreadPool
 import logging
 log = logging.getLogger(lib.settings.LOGGER)
 
-NUM_ENDPOINTS = 10
+NUM_ENDPOINTS = 12
 
 def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
 
@@ -141,6 +141,21 @@ def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
         except HTTPError as e:
             log.error((e.response.url, e.response.status_code))
             error_responses.append((e.response.url, e.response.status_code))
+
+    # PUT /{db}/{local-doc-id}
+    try:
+        doc = user.add_doc("_local/a-local-doc", content={"message": "I should not be replicated"})
+    except HTTPError as e:
+        log.error((e.response.url, e.response.status_code))
+        error_responses.append((e.response.url, e.response.status_code))
+
+    # GET /{db}/{local-doc-id}
+    try:
+        doc = user.get_doc("_local/a-local-doc")
+        assert(doc["content"]["message"] == "I should not be replicated")
+    except HTTPError as e:
+        log.error((e.response.url, e.response.status_code))
+        error_responses.append((e.response.url, e.response.status_code))
 
     # GET /{db}/_all_docs
     try:
--- a/functional_tests/test_db_online_offline.py
+++ b/functional_tests/test_db_online_offline.py
@@ -18,7 +18,7 @@ from multiprocessing.pool import ThreadPool
 import logging
 log = logging.getLogger(lib.settings.LOGGER)
 
-NUM_ENDPOINTS = 14
+NUM_ENDPOINTS = 13
 
 def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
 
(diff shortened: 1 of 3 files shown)


functional_tests/test_db_online_offline.py
```diff
@@ -18,7 +19,7 @@
 import logging
 log = logging.getLogger(lib.settings.LOGGER)
 
-NUM_ENDPOINTS = 10
+NUM_ENDPOINTS = 13
```
ashvindersingh: I think a better approach would be to have all the API in array or dictionary and then just use count to get the num rather than manually updating global variable.

functional_tests/test_db_online_offline.py
```diff
@@ -129,6 +123,14 @@ def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
     bulk_doc_errors = user.add_docs(num_docs=num_docs, bulk=True)
     error_responses.extend(bulk_doc_errors)
 
+    # POST /{db}/
```
ashvindersingh: This is a redundant check. We are already doing add_docs above.

functional_tests/test_db_online_offline.py
```diff
@@ -142,10 +144,37 @@ def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
...
 
+    # PUT /{db}/{local-doc-id}
+    local_doc_id = uuid.uuid4()
+    try:
+        doc = user.add_doc("_local/{}".format(local_doc_id), content={"message": "I should not be replicated"})
+    except HTTPError as e:
+        log.error((e.response.url, e.response.status_code))
+        error_responses.append((e.response.url, e.response.status_code))
+
+    # GET /{db}/{local-doc-id}
```
ashvindersingh: A check should be added where another user tries to replicate the same _local_doc id but it (another user) gets a error response.

functional_tests/test_db_online_offline.py
```diff
@@ -142,10 +144,37 @@ def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
...
+    except HTTPError as e:
+        log.error((e.response.url, e.response.status_code))
+        error_responses.append((e.response.url, e.response.status_code))
+
     # GET /{db}/_all_docs
     try:
         all_docs_result = user.get_all_docs()
-        assert len(all_docs_result["rows"]) == num_docs * 2
+        # num_docs /{db}/{doc} PUT + num_docs /{db}/_bulk_docs + num_docs POST /{db}/
+        assert(len(all_docs_result["rows"]) == num_docs * 3)
```
ashvindersingh: Can you please explain the multiplier with 3

functional_tests/test_db_online_offline.py
```diff
@@ -18,7 +19,7 @@
 import logging
 log = logging.getLogger(lib.settings.LOGGER)
 
-NUM_ENDPOINTS = 10
+NUM_ENDPOINTS = 13
 
 def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
```
ashvindersingh: I think applying this function again should yield the same result as the name applies. However the implementation doesn't support the above behavior.

functional_tests/test_db_online_offline.py
sethrosetter: Do you mean the following?

A scenario where user-1 puts _local/doc_1. Then user-2 puts _local/doc_1 to the same sync_gateway. Assert that there is a conflict error for user-2?

functional_tests/test_db_online_offline.py
sethrosetter: I'm not really sure what you mean by this. Can you please elaborate? There is some persistent state between test runs, however, it's pretty tied to these scenarios and should not be used like a library function.

functional_tests/test_db_online_offline.py
ashvindersingh: Yeah something similar. Actually user-2 should not be able to access the document and that should return error code if User-2 tries to access or modify it.

functional_tests/test_db_online_offline.py
ashvindersingh: The behavior of this should be idempotent. Test case should be able to use this function multiple times with true/false return value. or the signature should be changed to allow expected_num_docs/revisions etc.

PR #196, opened 2016-01-06
Fix "test_multiple_users_multiple_channels_mulitple_revisions" test

ashvindersingh:
Added logic to ignore the "_user" object in changes feed and returning total sum to not include '_user'.


ashvindersingh:
Thanks @tleyden @adamcfraser for the comments. In this particular case, this function is expected to return only the test framework generated docs found in changes feed. We should also hit scenarios where we encounter multiple "_user" objects in changes feed. opened #202

PR #199, opened 2016-01-06
Add ability to build with -race flag (or any other build flags)

tleyden:


tleyden:
Needed for https://github.com/couchbase/sync_gateway/issues/1400

sethrosetter:
@tleyden does this affect existing behavior in anyway if no flags are supplied?

tleyden:
It shouldn't, because it will just be passing an empty string to `build.sh` as an argument

sethrosetter:
@tleyden this crashes when no build flags are specified due to passing None

tleyden:
Ok great catch!  We just need a way to convert None to an empty string .. unless you have a cleaner way in mind.

sethrosetter:
@tleyden That should work

tleyden:
@sethrosetter can you take a look?

PR #203, opened 2016-01-07
Fixes #192

### LINKED ISSUE #192: Validate that cbgt pindexes are distributed evenly during / after performance test
Query the CBGT REST api `http://ec2-54-157-62-226.compute-1.amazonaws.com:4985/_cbgt/api/cfg` and verify that pindexes are spread evenly among nodes.

In this example, one writer node has all the pindexes assigned to it:

https://gist.github.com/tleyden/00af77d55c1feb3b99ed

This code can be re-used in performance tests and functional tests.

tleyden:
https://github.com/couchbaselabs/sync-gateway-testcluster/issues/192

Tested by running:

```
python performance_tests/run_tests.py --number-pullers 1000 --number-pushers 1000 --use-gateload --gen-gateload-config --reset-sync-gw --test-id="perftest"
```

I can see in the output:

```
PLAY RECAP ********************************************************************
sg1                        : ok=9    changed=3    unreachable=0    failed=0
sg2                        : ok=9    changed=3    unreachable=0    failed=0
sg3                        : ok=7    changed=2    unreachable=0    failed=0

>>> Detected valid CBGT Pindex distribution   <----------
Using Gateload
>>> Starting gateload with 1000 pullers and 1000 pushers
```

### DIFF
--- a/lib/cluster.py
+++ b/lib/cluster.py
@@ -125,7 +125,7 @@ class Cluster:
 
         # Validate CBGT
         if is_distributed_index:
-            if not self.validate_cbgt_pindex_distribution_retry(config):
+            if not self.validate_cbgt_pindex_distribution_retry():
                 self.save_cbgt_diagnostics()
                 raise Exception("Failed to validate CBGT Pindex distribution")
             print(">>> Detected valid CBGT Pindex distribution")
--- a/lib/cluster.py
+++ b/lib/cluster.py
@@ -55,7 +55,6 @@ class Cluster:
         if num_index_readers == 0:
             raise Exception("Functional tests require at least 1 index reader")
 
-    
     def reset(self, config):
 
         self.validate_cluster()
@@ -78,6 +77,7 @@ class Cluster:
         conf_path = os.path.abspath("conf/" + config)
         bucket_names_from_config = []
 
+        is_distributed_index = False
         with open(conf_path, "r") as config:
             data = config.read()
 
@@ -95,6 +95,7 @@ class Cluster:
 
             # Add CBGT buckets
             if "cluster_config" in conf_obj.keys():
+                is_distributed_index = True
                 bucket_names_from_config.append(conf_obj["cluster_config"]["bucket"])
 
             dbs = conf_obj["databases"]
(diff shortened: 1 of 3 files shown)


tleyden:
Just pushed up another version of this change which will dump out the CBGT cfg and diagnostics in the case of failure.

lib/cluster.py
```diff
@@ -120,6 +122,88 @@ def reset(self, config):
...
+        cbgt_diagnostics = adminApi.get_cbgt_diagnostics()
+        cbgt_cfg = adminApi.get_cbgt_cfg()
+        
+        # dump raw diagnostics
+        # (unfortunately this doesn't parse in as json)
+        print("CBGT diagnostic output: {}".format(cbgt_diagnostics))
+
+        # dump cfg json 
+        pretty_print_json = json.dumps(cbgt_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+        print("CBGT CFG output: {}".format(pretty_print_json))
```
sethrosetter: Perhaps we should dump to a file in case this happens during a run with lots of output?

lib/cluster.py
```diff
@@ -120,6 +122,88 @@ def reset(self, config):
...
+        any pindexes assigned to it
+        """
+        for i in xrange(10):
+            is_valid = self.validate_cbgt_pindex_distribution()
+            if is_valid:
+               return True
+            else:
+                print("Could not validate CBGT Pindex distribution.  Will retry after sleeping ..")
+                time.sleep(5)
+
```
sethrosetter: Return False explicity at the end of the method

sethrosetter:
If you are running in channel cache mode and call cluster.reset(), will this blow up?

lib/cluster.py
tleyden: Sure, where should I dump it?

lib/cluster.py
tleyden: Good call

lib/cluster.py
sethrosetter: /tmp/cbgt_error<time-stamp>.log ?

tleyden:
> If you are running in channel cache mode and call cluster.reset(), will this blow up?

Most likely, yes.   I will test it.

lib/cluster.py
```diff
@@ -120,6 +122,88 @@ def reset(self, config):
         )
         assert(status == 0)
 
+        # Validate CBGT
+        if not self.validate_cbgt_pindex_distribution_retry():
+            self.save_cbgt_diagnostics()
+            raise Exception("Failed to validate CBGT Pindex distribution")
+
+        print(">>> Detected valid CBGT Pindex distribution")
```
ashvindersingh: Can we use logging module instead of prints?

tleyden: @ashvindersingh the rest of the file is using `print`, so I will leave it as-is.  If we want to mass-convert the file to using the logging module, it should be in a separate PR.

ashvindersingh: ok, sure. I mentioned because half repo (functional tests) uses logging module while all the perf stuff is using prints. Going forward we should be consistent across repo.

tleyden:
Note to self: to test channel cache mode:

```
python provision/install_sync_gateway.py --branch master --config-file-path conf/sync_gateway_channel_cache.json
```

and then run any of the tests under test_overloaded_channel_cache.py

tleyden:
@sethrosetter can you test channel cache mode

PR #215, opened 2016-01-12
Silent Failure Audit

sethrosetter:
Remove try / excepts in futures where no exception should be throw. The existing behavior can cause tests to fail silently.

### DIFF
--- a/functional_tests/test_db_online_offline.py
+++ b/functional_tests/test_db_online_offline.py
@@ -340,7 +340,7 @@ def test_online_to_offline_changes_feed_controlled_close_continuous(cluster, con
                 docs_in_changes = future.result()
                 log.info("DOCS FROM CHANGES")
                 for k, v in docs_in_changes.items():
-                    log.info("DFC -> {}:{}".format(k, v))
+                    log.debug("DFC -> {}:{}".format(k, v))
 
     log.info("Number of docs from _changes ({})".format(len(docs_in_changes)))
     log.info("Number of docs add errors ({})".format(len(doc_add_errors)))
@@ -468,15 +468,15 @@ def test_online_to_offline_changes_feed_controlled_close_longpoll(cluster, conf,
                 try:
                     docs_in_changes = future.result()
                 except Exception as e:
-                    log.info(e)
-                    log.info("POLLING DONE EXCEPTION")
-                    log.info("AARGS: {}".format(e.args))
+                    log.warn(e)
+                    log.warn("POLLING DONE EXCEPTION")
+                    log.warn("ARGS: {}".format(e.args))
                     docs_in_changes = e.args[0]["docs"]
                     last_seq_num = e.args[0]["last_seq_num"]
-                    log.info("DOCS FROM longpoll")
+                    log.warn("DOCS FROM longpoll")
                     for k, v in docs_in_changes.items():
-                        log.info("DFC -> {}:{}".format(k, v))
-                    log.info("LAST_SEQ_NUM FROM longpoll {}".format(last_seq_num))
+                        log.debug("DFC -> {}:{}".format(k, v))
+                    log.warn("LAST_SEQ_NUM FROM longpoll {}".format(last_seq_num))
 
     log.info("Number of docs from _changes ({})".format(len(docs_in_changes)))
     log.info("last_seq_num _changes ({})".format(last_seq_num))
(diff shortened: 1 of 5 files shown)


sethrosetter:
@ashvindersingh I'm going to run against this branch in Jenkins and let you know when it is ready for review.

functional_tests/test_db_online_offline.py
```diff
@@ -327,24 +327,20 @@ def test_online_to_offline_changes_feed_controlled_close_continuous(cluster, con
...
+            if task_name == "db_offline_task":
+                log.info("DB OFFLINE")
+                # make sure db_offline returns 200
+                assert(future.result() == 200)
+            elif task_name == "docs_push":
+                log.info("DONE PUSHING DOCS")
+                doc_add_errors = future.result()
+            elif task_name == "continuous":
+                docs_in_changes = future.result()
+                log.info("DOCS FROM CHANGES")
```
ashvindersingh: You might want to change this to log.debug()

functional_tests/test_db_online_offline.py
```diff
@@ -327,24 +327,20 @@ def test_online_to_offline_changes_feed_controlled_close_continuous(cluster, con
...
-            except Exception as e:
-                print("Futures: error: {}".format(e))
+            task_name = futures[future]
+
+            if task_name == "db_offline_task":
+                log.info("DB OFFLINE")
+                # make sure db_offline returns 200
+                assert(future.result() == 200)
+            elif task_name == "docs_push":
+                log.info("DONE PUSHING DOCS")
```
ashvindersingh: log.debug() instead?

functional_tests/test_db_online_offline.py
```diff
@@ -461,34 +453,30 @@ def test_online_to_offline_changes_feed_controlled_close_longpoll(cluster, conf,
...
+            if task_name == "docs_push":
+                log.info("DONE PUSHING DOCS")
+                doc_add_errors = future.result()
+            if task_name == "polling":
+                # Long poll will exit with 503, return docs in the exception
+                log.info("POLLING DONE")
+                try:
+                    docs_in_changes = future.result()
+                except Exception as e:
+                    log.info(e)
```
ashvindersingh: In exception handling we should probably atleast have log.warn() or log.error()

functional_tests/test_db_online_offline.py
```diff
@@ -461,34 +453,30 @@ def test_online_to_offline_changes_feed_controlled_close_longpoll(cluster, conf,
...
+                    log.info("POLLING DONE EXCEPTION")
+                    log.info("AARGS: {}".format(e.args))
+                    docs_in_changes = e.args[0]["docs"]
+                    last_seq_num = e.args[0]["last_seq_num"]
+                    log.info("DOCS FROM longpoll")
+                    for k, v in docs_in_changes.items():
+                        log.info("DFC -> {}:{}".format(k, v))
+                    log.info("LAST_SEQ_NUM FROM longpoll {}".format(last_seq_num))
```
ashvindersingh: ditto as above comment for all the log statements in exception handling.

functional_tests/test_dcp_reshard.py
```diff
@@ -47,15 +47,12 @@ def test_dcp_reshard_sync_gateway_goes_down(cluster, conf):
...
-                    assert shutdown_status == 0
-                print("{} Completed:".format(tag))
-            except Exception as e:
-                print("Exception thrown while adding docs: {}".format(e))
+            tag = futures[future]
+            if tag == "sg_down":
+                # Assert takedown was successful
+                shutdown_status = future.result()
+                assert shutdown_status == 0
+            print("{} Completed:".format(tag))
```
ashvindersingh: The command never gets printed if the assert fails. This can be moved up before assert. Also if this can be changed to logging module.

functional_tests/test_dcp_reshard.py
```diff
@@ -102,22 +99,15 @@ def test_dcp_reshard_sync_gateway_comes_up(cluster, conf):
...
-                print("{} Completed:".format(tag))
-            except Exception as e:
-                print("Exception thrown while adding docs: {}".format(e))
-            else:
-                print "Docs added!!"
+            tag = futures[future]
+            if tag == "sg_up":
+                up_status = future.result()
+                assert up_status == 0
+            print("{} Completed:".format(tag))
```
ashvindersingh: same as above comment.

functional_tests/test_db_online_offline.py
sethrosetter: I would like to keep most of these as info, just so in the test output, I have an idea of what tasks completed. I will update the printing of docs to debug though.

functional_tests/test_db_online_offline.py
sethrosetter: Same as above.

functional_tests/test_db_online_offline.py
sethrosetter: In this test case, the exception is expected, so I am going with warn, because it is not technically an error

PR #216, opened 2016-01-13
now raising an exception with value error

ashvindersingh:
One liner but important change.


sethrosetter:
@ashvindersingh Can you kick of the functional tests targeting this branch and we can merge when the results come back?

ashvindersingh:
Yeah good point! In this particular case one test using the parallel function should validate the change. Going forward we should continue with this practice.

sethrosetter:
@ashvindersingh I completely agree.

sethrosetter:
@ashvindersingh Did you test this out / am I able to merge this?

PR #217, opened 2016-01-13
Use log.debug for user REST, change default back to log.info

sethrosetter:
- Use log.debug in most user method 
- Default logging back to log.info


ashvindersingh:
Looks good.

PR #224, opened 2016-01-15
Dual binary provisioning

### LINKED ISSUE #208: Update provisioning to use dual binaries
Once https://github.com/couchbase/sync_gateway/issues/1387 is completed, We need to update provisioning to deploy 2 binaries vs the shared binary.

sethrosetter:
Needs discussion:
- Bumping sleep in dcp reshard test
- Changing scenario parameters in continuous / longpoll tests

Fixes #208 

- Rips out unused CBGT bucket
- Install sg_accel for "sync_gateway_index_writers"
- Works for 
```
python provision/install_sync_gateway --version=1.2.0-46
python provision/install_sync_gateway --branch=master
```
- cluster.reset() will only launch sg_accel writers if the configuration provided is a distributed index conf
- cluster.verify_alive(mode) will check that sg_accels and sync_gateways are alive after a distribute index test and check that sync_gateways are alive for a channel cache test

Biggest changes
- cluster.sync_gateways are now only readers
- A few tests were operating in mixed mode for distributed index. All traffic is pointed at a reader for those scenarios now to enable compatibility with running in channel cache mode. Needs careful review to determine if changing these traffic endpoints will make a significant functional change
- Many provisioning changes to enable installing different packages to the different node types

Blocked on a clean jenkins run


tleyden:
Have you run anything on jenkins against this branch?

PR #230, opened 2016-01-15
todolite config

tleyden:
Fixes https://github.com/couchbaselabs/sync-gateway-testcluster/issues/229


sethrosetter:
@tleyden should this have the suffix "_di.json" and should there also be a channel cache version as well?

tleyden:
There are db's defined for both cc and di

sethrosetter:
@tleyden Didn't see that. whoops

PR #236, opened 2016-01-18
Add ability to skip bucket flush when installing SG

tleyden:
This was needed to retest https://github.com/couchbase/sync_gateway/issues/1475

### DIFF
--- a/provision/ansible/playbooks/install-sync-gateway-package.yml
+++ b/provision/ansible/playbooks/install-sync-gateway-package.yml
@@ -36,6 +36,7 @@
 
   tasks:
   - include: tasks/flush-server-buckets.yml
+    when: (skip_bucketflush != 'true') and (skip_bucketflush != 'True')
 
 # Create sync_gateway user
 - hosts: sync_gateways
--- a/provision/ansible/playbooks/install-sync-gateway-source.yml
+++ b/provision/ansible/playbooks/install-sync-gateway-source.yml
@@ -35,6 +35,7 @@
 
   tasks:
   - include: tasks/flush-server-buckets.yml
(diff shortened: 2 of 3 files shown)


provision/ansible/playbooks/install-sync-gateway-package.yml
```diff
@@ -36,7 +36,8 @@
 
   tasks:
   - include: tasks/flush-server-buckets.yml
-
+    when: skip_bucketflush == "false"
+
```
sethrosetter: Should this be true?

sethrosetter: Does a python False get correctly converted to "false" here? Just double checking.

tleyden: That looks right to me .. it only flushes server buckets if *skipping* bucketflush is false.  

I agree though it's a bit confusing.  

These variations would be less confusing:

```
   tasks:
    - include: tasks/flush-server-buckets.yml		    
      unless: skip_bucketflush == "true"
```

```
   tasks:
    - include: tasks/flush-server-buckets.yml		    
      when: do_bucketflush == "true"
```

tleyden: > Does a python False get correctly converted to "false" here? Just double checking.

It seemed to in my testing.  To be honest I was confused on how to do this comparison .. I'll try to switch to compare against False (boolean) and see what happens

tleyden:
Test results:

https://gist.github.com/tleyden/7ffab1941425ff286426

New commit will be pushed soon

tleyden:
@sethrosetter should be ready to merge now.

PR #238, opened 2016-01-19
Last minute functional tests

### LINKED ISSUE #231: Test compression with different user agent strings
Currently the test-bulk-compression only tests with empty user agent -- should test with user agent `CouchbaseLite/1.2` user agent and verify the behavior is in line with https://github.com/couchbase/sync_gateway/issues/1419#issuecomment-171483866

### LINKED ISSUE #235: Functional test: Support CouchbaseLite use of row-level sequences (as opposed to using last_seq)
See https://github.com/couchbase/sync_gateway/issues/1475#issuecomment-172426052

sethrosetter:
Fixes #235 
Fixes #231 

One question. Does it matter if I I use a channel cache vs distributed index config for the compression test or do both modes exercise the same code path?

### DIFF
--- a/functional_tests/test_seq.py
+++ b/functional_tests/test_seq.py
@@ -44,7 +44,7 @@ def test_seq(cluster, conf, num_users, num_docs, num_revisions):
     time.sleep(5)
 
     user_0_changes = users[0].get_changes(since=0)
-    doc_seq = user_0_changes["results"][250]["seq"]
+    doc_seq = user_0_changes["results"][num_docs / 2]["seq"]
 
     # https://github.com/couchbase/sync_gateway/issues/1475#issuecomment-172426052
     # verify you can issue _changes with since=12313-0::1023.15
--- a/functional_tests/test_seq.py
+++ b/functional_tests/test_seq.py
@@ -1,0 +1,65 @@
+import time
+
+import pytest
+
+
+from lib.admin import Admin
+from lib.verify import verify_changes
+
+import lib.settings
+import logging
+log = logging.getLogger(lib.settings.LOGGER)
+
+from fixtures import cluster
+
+
+@pytest.mark.distributed_index
+@pytest.mark.sanity
+@pytest.mark.parametrize(
+        "conf, num_users, num_docs, num_revisions", [
+            ("sync_gateway_default_functional_tests_di.json", 10, 500, 1),
+            ("sync_gateway_default_functional_tests_cc.json", 10, 500, 1)
+        ],
+        ids=["DI-1", "CC-2"]
+)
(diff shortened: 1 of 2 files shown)


functional_tests/test_seq.py
```diff
@@ -0,0 +1,81 @@
...
+    for user in users:
+        user.add_docs(num_docs, bulk=True)
+
+    for user in users:
+        user.update_docs(num_revisions)
+
+    time.sleep(5)
+
+    user_0_changes = users[0].get_changes(since=0)
+    doc_seq = user_0_changes["results"][250]["seq"]
```
tleyden: Instead of hardcoding 250, could you do `num_docs / 2`?

functional_tests/test_seq.py
```diff
@@ -0,0 +1,81 @@
...
+        last_doc_entry_seq = changes["results"][-1]["seq"]
+
+        log.info('Second to last doc "seq": {}'.format(second_to_last_doc_entry_seq))
+        log.info('Last doc "seq": {}'.format(last_doc_entry_seq))
+
+        if mode == "distributed_index":
+            # Verify last "seq" follows the formate 12313-0, not 12313-0::1023.15
+            log.info('Verify that the last "seq" is a plain hashed value')
+            assert(len(second_to_last_doc_entry_seq.split("::")) == 2)
+            assert(len(last_doc_entry_seq.split("::")) == 1)
```
tleyden: rather than checking the format, which might change in future releases, why not just try to issue a _changes request with the second_to_last_doc_entry_seq as the since value?  on anything before 1.2.0-50, this should produce a 500 error and the test will fail.

sethrosetter: Yes. Line 52 does issue a since with the format that was currently breaking.  https://github.com/couchbaselabs/sync-gateway-testcluster/pull/238#discussion-diff-50153268R52. I added the extra 'seq' formatting check per @adamcfraser's request last night.

adamcfraser: I agree with Traun about not relying on the format.  I think the intended validation of a 'plain hashed value' here is to ensure we don't have an infinite loop scenario - where subsequent requests for the last sequence's since value return data, including that same sequence value.

Instead of using the format for this, you could check the response of the additional _changes response that Traun suggested, and ensure there aren't any docs returned.

adamcfraser: Apologies for not suggesting this approach in the first place @sethrosetter.

tleyden: > Yes. Line 52 does issue a since with the format that was breaking breaking in 1.2.0-50

Ok, I didn't see that.  

I think the extra formatting stuff probably won't hurt, and we can investigate future errors as we see them.

Going to merge as-is.

PR #248, opened 2016-01-20
Use POST _changes rather than GET

### LINKED ISSUE #240: Use POST _changes instead of GET

sethrosetter:
Waiting on no regressions - http://qa.sc.couchbase.com/view/mobile/job/centos-syncgateway-functional-tests/105/

Fixes #240 

- Convert existing _changes requests to use POST instead of GET
- Update longpoll implementation to use parameters closer to what the client would send (Android)


sethrosetter:
@ashvindersingh this should be good to go.

ashvindersingh:
@sethrosetter look good!

PR #265, opened 2016-01-26
Bump ansible async timeout, bump collection interval, use go 1.5.3

sethrosetter:


sethrosetter:
Needs a slight refactor before it can go in.

tleyden:
Just noticed this PR as the only PR outstanding.  If it's adding value, should we merge and refactor as a separate PR?

sethrosetter:
@tleyden This should be good to go now

PR #280, opened 2016-02-11
User newer couchbase server and sync gateway in example commands

tleyden:
Also mention that at least 2 SG's are required


sethrosetter:
@tleyden Thanks!

PR #293, opened 2016-02-23
Change readme to use OSX by default

tleyden:
I think more people are going to run this on OSX than docker, so move the docker stuff to a wiki page and default to OSX


sethrosetter:
@tleyden Thanks!

PR #296, opened 2016-02-23
Make detected_pattern function more robust against errors

### LINKED ISSUE #295: detected_pattern function not robust
if detected_pattern is called with a zip_file_path of None, it will currently throw an exception.

tleyden:
Fixes #295

### DIFF
--- a/functional_tests/fixtures.py
+++ b/functional_tests/fixtures.py
@@ -10,6 +10,7 @@ import settings
 from subprocess import Popen, PIPE
 import os.path
 
+import sys
 import logging
 import lib.settings
 log = logging.getLogger(lib.settings.LOGGER)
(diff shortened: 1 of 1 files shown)


functional_tests/fixtures.py
```diff
@@ -58,16 +59,24 @@ def detected_panics(zip_file_path):
...
+    
+    try:
+        if not os.path.isfile(zip_file_path):
+            log.error("Can't run zipgrep, cannot find zipfile: {}".format(zip_file_path))
+            return False
+
+        process = Popen(["zipgrep", pattern, zip_file_path], stdout=PIPE)
+        (output, err) = process.communicate()
+        exit_code = process.wait()
+        if exit_code == 0:
```
sethrosetter: you could simplify these line with 
```
exit_code = subprocess.call(["zipgrep", pattern, zip_file_path])
if exit_code == 0:
    ...
```

tleyden: I don't think that will work because it won't capture the output, which is needed here: `log.info("Detected pattern {}: {}".format(pattern, output))`

PR #298, opened 2016-02-23
Fixes #297 Modify script to create buckets based on bucket shadowing …

### LINKED ISSUE #297: Modify script to create buckets based on bucket shadowing config
Also, add trivial functional test that simply sets up a bucket shadowing config.

tleyden:
…config

Changes:

- Refactor out `discover_buckets_from_config`
- Refactor out `discover_mode`
- Update `discover_buckets_from_config` to consider shadow buckets

### DIFF
--- a/lib/cluster.py
+++ b/lib/cluster.py
@@ -14,6 +14,7 @@ from lib.syncgateway import SyncGateway
 from lib.sgaccel import SgAccel
 from lib.server import Server
 from lib.admin import Admin
+from lib.config import Config
 from lib import settings
 from provision.ansible_runner import run_ansible_playbook
 
@@ -45,7 +46,8 @@ class Cluster:
         self.sg_accels = [SgAccel(sgw) for sgw in sgsw]
         self.servers = [Server(cb) for cb in cbs]
         self.load_generators = lds
-
+        self.sync_gateway_config = None  # will be set to Config object when reset() called
+        
     def _hosts_for_tag(self, tag):
         hostfile = "provisioning_config"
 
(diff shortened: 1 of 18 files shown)


lib/cluster.py
```diff
@@ -300,3 +272,81 @@ def __repr__(self):
...
+        if "cluster_config" in conf_obj.keys():
+            return  "distributed_index"
+
+    return "channel_cache"
+    
+
+def strip_invalid_json_from_config(data):
+
+    # strip out templated variables {{ ... }}
+        data = re.sub("({{.*}})", "0", data)
```
sethrosetter: @tleyden Indentation looks a little strange here

tleyden: oh, it's probably a tabs / spaces thing.  I'll fix it, thanks for catching.

lib/cluster.py
```diff
@@ -300,3 +272,81 @@ def __repr__(self):
...
+    # Buckets may be shared for different functionality
+    bucket_name_set = list(set(bucket_names_from_config))
+
+    return bucket_name_set
+
+    
+if __name__ == "__main__":
+
+    # poor man's unit test
+    conf_path = "conf/sync_gateway_bucketshadow_cc.json"
```
sethrosetter: I think we should do this a bit different. Maybe we could return mode and buckets from the reset method. We may need to to that in the future anyway for config testing.

tleyden: What about something like this?

```
config_wrapper = ConfigWrapper(conf_path)
for bucket in config_wrapper.buckets:
    ...
if config_wrapper.mode == DIST_INDEX:
    ...
if len(config_wrapper.num_databases) >1:
    ...

```

lib/cluster.py
tleyden: wasn't a tab / spaces thing, just needed some more indents.  committed a fix locally.

tleyden:
@sethrosetter I re-pushed with the changes we discussed -- take a look.

```
py.test -s "functional_tests/test_bucket_shadow.py"
```

is passing, but I have not tried to retest the other tests

tleyden:
Updated PR to remove the `test_bucket_shadow.py` file, since that is still in progress

sethrosetter:
Test run succeeded.

PR #310, opened 2016-03-01
Fixes #308 limitation when deleting / recreating a single bucket

### LINKED ISSUE #308: Fix limitation when deleting / recreating a single bucket
Problem: if you create two buckets initially, then delete one bucket and recreate it, it will try to recreate the second bucket with the full RAM of the cluster but fail (since one bucket is still using half of the RAM)

tleyden:
Tested via the following:

Add this snippet to `test_bucket_shadow.py`:

```
def test_create_bucket(cluster):
    
    bucket = cluster.servers[0].create_buckets(["foo", "bar"])
    print "Created bucket: {}".format(bucket)
```

and then kick off test via:

```
py.test -s functional_tests/test_bucket_shadow.py::test_create_bucket
```

Expected log output:

https://gist.github.com/tleyden/bb5e8bc975728384894f

Then delete bar from the couchbase web ui, and run:

```
def test_create_bucket(cluster):
    
    bucket = cluster.servers[0].create_buckets(["bar"])
    print "Created bucket: {}".format(bucket)
```

It will recreate it.  Log output: https://gist.github.com/tleyden/48ecea1e3e045b985051

### DIFF
--- a/conf/sync_gateway_bucketshadow_low_revs_cc.json
+++ b/conf/sync_gateway_bucketshadow_low_revs_cc.json
@@ -4,13 +4,14 @@
     "maxIncomingConnections": 0,
     "maxCouchbaseConnections": 16,
     "maxFileDescriptors": 90000,
-     "slowServerCallWarningThreshold": 500,
+    "slowServerCallWarningThreshold": 500,
     "compressResponses": true,
     "log": ["*"],
     "databases":{
         "db":{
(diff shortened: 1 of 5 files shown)


sethrosetter:
@tleyden This branch broke when running in Jenkins. http://qa.sc.couchbase.com/view/mobile/job/centos-syncgateway-functional-tests/145/ Can you try to run a subset locally? One fixture with a few tests should work.

PR #313, opened 2016-03-02
Add tests (currently passing) described in https://github.com/couchba…

tleyden:
https://github.com/couchbaselabs/sync-gateway-testcluster/issues/291#issuecomment-190858093

Even though these are all passing, they should serve as useful regression tests for bucket shadowing.  Would also like to run these overnight in a loop on Jenkins.


sethrosetter:
Should these be removed or automated - https://github.com/couchbaselabs/sync-gateway-testcluster/blob/feature/issue_291_additional_ftests/functional_tests/test_bucket_shadow.py#L179-L185 ?

tleyden:
Related to https://github.com/couchbase/sync_gateway/issues/1635

PR #335, opened 2016-03-09
Use admin port when checking to see if sync gateway up

tleyden:
Fixes this issue: https://github.com/couchbase/sync_gateway/pull/1636#issuecomment-194560277


tleyden:
Nevermind, looks like this causes more problems than it solves

PR #362, opened 2016-03-30
Pull jenkins results. Work around until jenkins can host the html pages

sethrosetter:


tleyden:
Is there any other way aside from hardcoding the ip addresses?

PR #367, opened 2016-03-31
Add dump-release-packages.py script

sethrosetter:
Add script to download and exact all mobile packages and write the package contents to file


tleyden:
Can you add a comment to the top of the file to this effect:

> This script will pull the packages defined below and save the package contents in a series of files:

Also, "dump" is a bit confusing.  How about renaming to download-release-packages.py

sethrosetter:
@tleyden Made the name change. Merging get unblocked on some work

PR #408, opened 2016-04-19
Add test_replication_config

tleyden:
and enhance the basic test to check the result of kicking off one shot replication

### DIFF
--- a/testsuites/syncgateway/functional/test_sg_replicate.py
+++ b/testsuites/syncgateway/functional/test_sg_replicate.py
@@ -13,12 +13,12 @@ import time
 
 DB1 = "db1"
 DB2 = "db2"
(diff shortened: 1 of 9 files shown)


testsuites/syncgateway/functional/test_sg_replicate.py
```diff
@@ -294,10 +305,20 @@ def test_stop_replication_via_replication_id():
     assert len(active_tasks) == 0
 
 
-def create_sync_gateways():
+def test_replication_config():
+
+    sg1, sg2 = create_sync_gateways("resources/sync_gateway_configs/sync_gateway_sg_replicate_continuous_cc.json")
+
+    # Wait until active_tasks is non empty
+    wait_until_active_tasks_non_empty(sg1)
```
sethrosetter: will this loop forever or throw an Exception after a period of time?

PR #419, opened 2016-04-21
Feature/sync gateway sanity nightly tagging

sethrosetter:
- Add nightly tag to tests taking longer than 3 min
- Add template for robot framework tests


_queue/template.robot
```diff
@@ -0,0 +1,48 @@
+*** Settings ***
```
tleyden: what is the meaning of the directory name `_queue`?

sethrosetter: I would like to provide a place for dev to put robot testcase definition to allow me to triage them and implement them. Maybe it needs a better name?

tleyden: Ok, makes sense.  I think queue is a bit confusing, what about calling it:

- `templates`
- `unimplemented`
- `in_progress`
- `disabled`

sethrosetter: I like `unimplemented`. I will change it.

sethrosetter:
Fixed

PR #421, opened 2016-04-21
Support dev builds in install_couchbase_server.py

sethrosetter:
Add support to target dev builds of Couchbase Server

### DIFF
--- a/libraries/provision/install_couchbase_server.py
+++ b/libraries/provision/install_couchbase_server.py
@@ -9,62 +9,69 @@ class CouchbaseServerConfig:
 
     def __init__(self, version):
 
-        self._valid_versions = [
+        version_build = version.split("-")
+        self.version = version_build[0]
+        if len(version_build) == 2:
+            # Build number is included
+            self.build = version_build[1]
+        else:
+            self.build = None
+
+        self.released_versions = [
             "3.1.0",
             "3.1.1",
             "3.1.2",
+            "3.1.3",
+            "3.1.4",
+            "3.1.5",
             "4.0.0",
             "4.1.0"
         ]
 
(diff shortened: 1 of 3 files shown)


libraries/provision/install_couchbase_server.py
```diff
@@ -9,62 +9,69 @@ class CouchbaseServerConfig:
...
-            print "Server package url not found. Make sure to specify a version / build."
-            raise ValueError("Invalid server url")
 
-    @property
-    def version(self):
-        return self._version
+            # Get dev server package from latestbuilds
+            if self.version.startswith("4.1"):
+                # http://172.23.120.24/builds/latestbuilds/couchbase-server/sherlock/5914/couchbase-server-enterprise-4.1.1-5914-centos7.x86_64.rpm
+                base_url = "http://172.23.120.24/builds/latestbuilds/couchbase-server/sherlock/{}".format(self.build)
```
tleyden: is there a dns name associated w/ 172.23.120.24?

tleyden: also, will this work on aws?

sethrosetter: > is there a dns name associated w/ 172.23.120.24?

I dont believe so. 



> also, will this work on aws?

Released versions (i.e. 4.1.0) will still look to s3 but dev builds (4.5.0-XXX, any version with a build number) of server will not work in AWS due to being behind VPN. This should not block perf or most mobile testing purposes for the time being as long as you depend on a released version.

tleyden: Can you update to cbnas01.sc.couchbase.com?

PR #585, opened 2016-07-07
Feature/readme update

sethrosetter:
- Update readme
- remove assertion that was causing a test to fail. When initiating a replication, it is possible to have a delay before the replication shows up in active tasks. Instead of failing immediately if a replication is not found, it will now poll.

---
This change is [Reviewable](https://reviewable.io/reviews/couchbaselabs/mobile-testkit/585)

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,50 +298,86 @@ TODO
 Listener Tests
 ==============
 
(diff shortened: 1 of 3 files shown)


tleyden:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/couchbaselabs/mobile-testkit/585)*

PR #616, opened 2016-08-08
Feature/rebalance support

### LINKED ISSUE #243: Functional test for rebalance scenario
Regression functional test to verify scenario https://github.com/couchbase/sync_gateway/issues/1286

sethrosetter:
DO NOT MERGE until Outstanding Tasks are checked off. 

- Add support for `rebalance_in` and `rebalance_out`
- Fixes #243 
- Add auth support for `verify_docs_in_changes`

### Outstanding tasks
- [ ] Implement worker approach for doc updates
- [ ] Add / Update doc load during rebalance in

---
This change is [Reviewable](https://reviewable.io/reviews/couchbaselabs/mobile-testkit/616)


keywords/CouchbaseServer.py
```diff
@@ -269,4 +260,158 @@ def get_server_docs_with_prefix(self, url, bucket, prefix):
...
+
+        return True
+
+    def rebalance_in(self, admin_server, server_to_add):
+        """
+        Adds a server from a pool and waits for rebalance to complete.
+        admin_server format:    http://localhost:8091
+        server_to_add format:   http://localhost:8091
+        """
+
```
adamcfraser: Are you taking server_to_add in the http://...:8091 format because that's the format you expect callers to already have available?  Otherwise it seems like you could just format as hostname.

testsuites/syncgateway/functional/1sg_1ac_2cbs/rebalance_scenarios.py
```diff
@@ -0,0 +1,61 @@
...
+    with ThreadPoolExecutor(5) as executor:
+
+        # Add docs to sg
+        log_info("Adding docs to sync_gateway")
+        add_docs_task = executor.submit(client.add_docs, sg_one_url, sg_db, num_docs, "test_doc", channels=channels, auth=session)
+        docs = add_docs_task.result()
+
+        # Start updating docs and rebalance out one CBS node
+        log_info("Updating docs on sync_gateway")
+        update_docs_task = executor.submit(client.update_docs, sg_one_url, sg_db, docs, num_updates, auth=session)
```
adamcfraser: How do you validate that the update docs task runs long enough for the rebalance to complete?  Just based on num_updates?  That's probably reasonable - it minimizes additional work -  but would be interesting to have a more worker-like approach where you can issue 'start updating' here, and then issue a 'stop updating' command after your rebalance is complete.  That would avoid any race scenarios where the updates finish mid-rebalance.

testsuites/syncgateway/functional/1sg_1ac_2cbs/rebalance_scenarios.py
```diff
@@ -0,0 +1,61 @@
...
+        log_info(updated_docs)
+
+    # Verify docs / revisions present
+    client.verify_docs_present(sg_one_url, sg_db, updated_docs, auth=session)
+
+    # Verify docs revisions in changes feed
+    client.verify_docs_in_changes(sg_one_url, sg_db, updated_docs, auth=session)
+
+    # Rebalance Server back in to the pool
+    assert server.rebalance_in(admin_server=cbs_one_url, server_to_add=cbs_two_url), "Could not rebalance node back in .."
```
adamcfraser: Would we want to verify rebalance_in under load as well?  (i.e. be writing docs while the rebalance in is happening)

keywords/CouchbaseServer.py
sethrosetter: This is what the callers have available in the when getting the cluster information in the test setup. This is kind of an edge case in the sense that the Couchbase Server Rest API has different formatting when passing the host ips in the POST params

adamcfraser: Fair enough - seems like a reasonable approach then.

testsuites/syncgateway/functional/1sg_1ac_2cbs/rebalance_scenarios.py
sethrosetter: That is a great idea. It is pretty add hoc at the moment. I will look into using this approach. This is not a time sensitive PR so I will wait on merging until I get a chance to try that.

testsuites/syncgateway/functional/1sg_1ac_2cbs/rebalance_scenarios.py
sethrosetter: The idea with this rebalance_in is that the state of the cluster needs to be consistent with topology that the tests are expecting (i.e. 1 sync_gateway, 1 accel, 2 Couchbase Servers). Since a server is rebalanced out as part of the test, only 1 server will be available for subsequent tests. Although since this test takes a long time (~12 min locally), I think it would be smart to get the most that we can out of this. I will add that using the same worker-like proposition above.

sethrosetter:
Will be reopened when ready

PR #735, opened 2016-10-12
Update requirements.txt

borrrden:
Add in an explicit dep for pip to avoid the cryptography pip package compilation madness

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,6 +24,7 @@ MarkupSafe==0.23
 mccabe==0.5.2
 netifaces==0.10.4
 paramiko==2.0.2
+pip==8.1.2
 py==1.4.31
 pyasn1==0.1.9
 pycodestyle==2.0.0


tleyden:
This feels odd to me since I thought pip was installing these requirements in the first place

tleyden:
@borrrden do you have any more details on the cryptography pip package compilation madness?  It feels weird to me to `pip install pip`, and I think we might be better off in the long run if we can avoid it.  Or maybe that's an accepted practice with `virtualenv`, in which case nevermind and we should just merge this.

borrrden:
The goal is to get a certain version of pip as the minimum. This was the command that was recommended by pip itself.

borrrden:
This was actually solved by @sethrosetter 's suggestion of running `pip install --upgrade virtualenv` from outside the virtual environment

sethrosetter:
@borrrden Can I close this then?

borrrden:
Yes, allow me.

PR #753, opened 2016-10-18
Add server install unit tests

sethrosetter:


framework_tests/test_couchbase_server_install.py
```diff
@@ -0,0 +1,35 @@
...
+    assert server_config.version == "4.1.1"
+    assert server_config.build == "5914"
+
+
+def test_enterprise():
+    server_config = CouchbaseServerConfig(
+        version="4.1.1-5914"
+    )
+    base_url, package_name = server_config.get_baseurl_package()
+    assert base_url == "http://cbnas01.sc.couchbase.com/builds/latestbuilds/couchbase-server/sherlock/5914"
```
tleyden: this feels like it has a dependency on a hardcoded value.  

what if it just asserted that the base_url had 5914 somewhere in it?  likewise, assert that `package_name` had 4.1.1 somewhere in the string?

sethrosetter: We depend on a hardcoded url value in the provisioning code so if that changes we could just change the to reflect the new download url? I really have no strong feelings either way. I just feel like this is more explicit

sethrosetter:
```
 assert base_url  contains  "couchbase.com"
asster package_name contains server_config.version (not as important)

 assert base_url  contains "s3.amazonaws.com"
```

PR #761, opened 2016-10-24
Add log scanner / fix webhook ip generation

### LINKED ISSUE #707: Improve log panic detection
- run panic analysis regardless of test failure status
- use grep instead of zipgrep

### LINKED ISSUE #724: Investigate webhook ip generation
error is in generate_cluster_from_pools.py

```
Failed to find local_ip, webhook tests will fail.  Error: [Errno 8] nodename nor servname provided, or not known
```

sethrosetter:
Tracking ticket to use log scanner everywhere - https://github.com/couchbaselabs/mobile-testkit/issues/762

Fixes #707 
Fixes #724


utilities/scan_logs.py
```diff
@@ -0,0 +1,30 @@
...
+    'error_strings' should be a list. Example ['panic', 'error']
+    """
+
+    if type(error_strings) != list:
+        raise ValueError("'error_strings must be a list'")
+
+    log_info("Looking for {} in {} ...".format(error_strings, log_file_path))
+
+    # Scan each line in the log file for the words to search for
+    with open(log_file_path) as f:
```
tleyden: This might blow up if the log file is huge, because I believe it will need to load the entire file contents in memory.  Using a [generator approach](http://www.dabeaz.com/generators/Generators.pdf) might be way more efficient here.

Any chance you can test on a 1 GB logfile and make sure that it can process it quickly?  If not, then I think we should consider switching to generators.

libraries/utilities/generate_clusters_from_pool.py
```diff
@@ -156,17 +160,22 @@ def write_config(config, pool_file):
         try:
             f.write("[webhook_ip]\n")
             if platform.system() == "Darwin":
-                local_ip = socket.gethostbyname(socket.gethostname())
+                # HACK: http://stackoverflow.com/questions/166506/finding-local-ip-addresses-using-pythons-stdlib
```
tleyden: Glad to see the hack documented, but can you make it clear what's being done here?  "Find the IP address of the machine this code is running on on the LAN so that other machines on the LAN can connect to it" <-- ?

utilities/scan_logs.py
tleyden: OK cool!

PR #767, opened 2016-10-27
Remove route53 dns setup

jamiltz:


sethrosetter:
@jamiltz Why is this necessary?

jamiltz:
That was to deploy a 1SG and 1CBS setup to my personal AWS account which had no pre configured route53. Made the change with @tleyden.

jamiltz:
I think we gave up on deploying to my personal account in the end but that was the intention with this PR. Hopefully it doesn't brake anything :| I'll try it again and report the results in https://github.com/couchbaselabs/mobile-testkit/issues/685

sethrosetter:
Sweet! Thanks. Talked about it with @tleyden and it shouldn't hurt anything.

PR #769, opened 2016-10-27
Give ability to generate arbitrary cluster config

tleyden:
Instead of this script having a fixed number of configs, take parameters that allow it to take an arbitrary config and it will add it to the fixed configs

Useful if people want to provision arbitrary sized clusters

### DIFF
--- a/libraries/utilities/generate_clusters_from_pool.py
+++ b/libraries/utilities/generate_clusters_from_pool.py
@@ -29,6 +29,15 @@ class ClusterDef:
             self.num_lbs
         )
 
+    def set_name_from_vars(self):
+        self.name = "{}sg_{}ac_{}cbs_{}lgs_{}lbs".format(
+            self.num_sgs,
+            self.num_acs,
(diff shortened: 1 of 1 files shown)


sethrosetter:
Is there a README enhancement needed?

sethrosetter:
What happens if you spec a custom config that already exists?

libraries/utilities/generate_clusters_from_pool.py
```diff
@@ -245,8 +271,28 @@ def generate_clusters_from_pool(pool_file):
...
+                      help="number of sg accel instances")
+
+    parser.add_option("", "--num-lbs",
+                      action="store", type="int", dest="num_lbs", default=0,
+                      help="number of load balancer instances")
+
+    parser.add_option("", "--num-lgs",
+                      action="store", type="int", dest="num_lgs", default=0,
+                      help="number of load generator instances")
+
```
sethrosetter: Provide validation? Are all required? Should they be?

PR #783, opened 2016-11-16
Feature/issue 782 erroneously build sg accel

### LINKED ISSUE #782: provision_cluster.py --sync-gateway-commit tries to build sg-accel even though it doesn't need to
To reproduce;

```
$ export CLUSTER_CONFIG=resources/cluster_configs/2sg_1cbs_1lgs
$ python libraries/provision/provision_cluster.py --server-version=4.5.0 --sync-gateway-commit="5aac91c8b98252449336b01aa392d6c3001d39a9"
```

and it's getting permanently stuck on:

```
TASK [SG ACCEL | Run bootstrap.sh script] **************************************
```

After sshing into the vagrant machine and manually running `bootstrap.sh`, I'm seeing this error:

```
The authenticity of host 'github.com (192.30.253.113)' can't be established.
RSA key fingerprint is 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)? The authenticity of host 'github.com (192.30.253.113)' can't be established.
RSA key fingerprint is 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)?
```

it’s happening `install-sync-gateway-source.yml` is calling `bootstrap.sh` with `-p sg-accel`, which causes it to try to clone the repo via ssh.

In my case I don't even have a valid ssh key on the vagrant machines, so even when I say `yes` I get `Permission denied (publickey).`

In any case, since my CLUSTER_CONFIG doesn't have any sg accels, it should only be trying to build sync gateway.  I've got a fix in progress.

tleyden:
Fixes #782

### DIFF
--- a/libraries/provision/ansible/playbooks/install-common-tools.yml
+++ b/libraries/provision/ansible/playbooks/install-common-tools.yml
@@ -93,5 +93,5 @@
   - name: DEPS | install repo tool
     get_url: url=https://storage.googleapis.com/git-repo-downloads/repo dest=/usr/bin/repo mode=0751
 
-  - name: DEPS | disable git color UI config
-    shell: git config --global color.ui false
+
+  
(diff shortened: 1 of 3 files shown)


tleyden:
@sethrosetter should be ready for merge

PR #826, opened 2016-12-19
fix linting issues / import

sethrosetter:
fix import issue in RemoteExecutor class


sethrosetter:
Not for this sort of change. A review should be fine. Areas where we touch framework code or tests, we should run sanity jobs against the branch

PR #829, opened 2016-12-20
Set the revs limit on the perf test

tleyden:
Limits docs from growing too large and skewing results


sethrosetter:
Can you update the name of the config to reflect this change?

PR #832, opened 2016-12-21
Adding the changes for vagrant ethernet public network

raghusarangapani:
The vagrant file under public_network assumes en0 wifi as the connection type. It does not work for Thunderbolt ethernet connections. This change adds support to Thunderbolt ethernet connection.

generate_cluster_configs_from_vagrant_hosts.py was also modified to accommodate this new vagrant file.

### DIFF
--- a/utilities/generate_cluster_configs_from_vagrant_hosts.py
+++ b/utilities/generate_cluster_configs_from_vagrant_hosts.py
@@ -17,8 +17,10 @@ def generate_cluster_configs_from_vagrant(private_network, public_network, publi
     3. Uses this IP list to build a pool.json file and generate the cluster configurations
     """
 
-    if private_network and public_network and public_network_ethernet:
-        raise ProvisioningError("Invalid private_network, public_network and public_network_ethernet flags")
+    if private_network and public_network:
+        raise ProvisioningError("Invalid private_network and public_network flags")
+    elif private_network and public_network_ethernet:
+        raise ProvisioningError("Invalid private_network and public_network_ethernet flags")
 
     if not private_network and not public_network and not public_network_ethernet:
         raise ProvisioningError("Invalid private_network, public_network and public_network_ethernet flags")
--- a/utilities/generate_cluster_configs_from_vagrant_hosts.py
+++ b/utilities/generate_cluster_configs_from_vagrant_hosts.py
@@ -10,26 +10,28 @@ import os
 from libraries.utilities.generate_clusters_from_pool import generate_clusters_from_pool
 
 
-def generate_cluster_configs_from_vagrant(private_network, public_network):
+def generate_cluster_configs_from_vagrant(private_network, public_network, public_network_ethernet):
     """
     1. Gets the status for a running vagrant vm set.
     2. Uses the host name to look up the ip allocated to each vagrant vm instance
     3. Uses this IP list to build a pool.json file and generate the cluster configurations
     """
 
-    if private_network and public_network:
-        raise ProvisioningError("Invalid private_network and public_network flags")
+    if private_network and public_network and public_network_ethernet:
+        raise ProvisioningError("Invalid private_network, public_network and public_network_ethernet flags")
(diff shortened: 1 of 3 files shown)


utilities/generate_cluster_configs_from_vagrant_hosts.py
```diff
@@ -10,26 +10,28 @@
...
+def generate_cluster_configs_from_vagrant(private_network, public_network, public_network_ethernet):
     """
     1. Gets the status for a running vagrant vm set.
     2. Uses the host name to look up the ip allocated to each vagrant vm instance
     3. Uses this IP list to build a pool.json file and generate the cluster configurations
     """
 
-    if private_network and public_network:
-        raise ProvisioningError("Invalid private_network and public_network flags")
+    if private_network and public_network and public_network_ethernet:
```
sethrosetter: Add a different type of verification. What were are looking for is that only one flag is set and the other two are not. You could use a count to make sure only one is set or use a series of asserts.

vagrant/public_network_ethernet/Vagrantfile
```diff
@@ -0,0 +1,33 @@
...
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure(2) do |config|
+
+  host_ips = [
+    "host1",
+    "host2",
+    "host3",
+    "host4"
```
sethrosetter: Add a host5. You can see that this is the setting by default on master

utilities/generate_cluster_configs_from_vagrant_hosts.py
raghusarangapani: Added an else clause to keep it simple

vagrant/public_network_ethernet/Vagrantfile
raghusarangapani: Added

utilities/generate_cluster_configs_from_vagrant_hosts.py
```diff
@@ -19,17 +19,21 @@ def generate_cluster_configs_from_vagrant(private_network, public_network):
 
     if private_network and public_network:
         raise ProvisioningError("Invalid private_network and public_network flags")
+    elif private_network and public_network_ethernet:
+        raise ProvisioningError("Invalid private_network and public_network_ethernet flags")
```
sethrosetter: Would public_network and public_network_ethernet still fall through?

I would check if one flag is set, the other two are not.

raghusarangapani: Fixed

sethrosetter:
Can you run `flake8` from the root of the repo and fix the following errors?
```
./utilities/generate_cluster_configs_from_vagrant_hosts.py:36:1: E101 indentation contains mixed spaces and tabs
./utilities/generate_cluster_configs_from_vagrant_hosts.py:36:1: W191 indentation contains tabs
./utilities/generate_cluster_configs_from_vagrant_hosts.py:38:1: E101 indentation contains mixed spaces and tabs
./utilities/generate_cluster_configs_from_vagrant_hosts.py:96:1: W293 blank line contains whitespace
```

utilities/generate_cluster_configs_from_vagrant_hosts.py
```diff
@@ -10,26 +10,22 @@
...
     2. Uses the host name to look up the ip allocated to each vagrant vm instance
     3. Uses this IP list to build a pool.json file and generate the cluster configurations
     """
 
-    if private_network and public_network:
-        raise ProvisioningError("Invalid private_network and public_network flags")
-
-    if not private_network and not public_network:
-        raise ProvisioningError("Invalid private_network and public_network flags")
-
```
sethrosetter: Why did you move this from the method into main? If someone was importing / calling this from another script, this validation would not longer be run.

PR #835, opened 2016-12-22
Feature/launch without num shards test

### LINKED ISSUE #74: Add test for sync_gateway launching without NUM_SHARDS
need funcitonal test for #1298 in sync_gateway

sethrosetter:
Fixes #74

### DIFF
--- a/testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
+++ b/testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
@@ -451,3 +451,40 @@ def test_take_all_sgaccels_down(params_from_base_test_setup, sg_conf):
     log_info("Verifying all the changes show up for 'a_user' ...")
     all_docs = doc_push_result_1 + docs_push_result_2 + doc_push_result_3
     client.verify_docs_in_changes(url=sg_url, db=sg_db, expected_docs=all_docs, auth=a_user_session, polling_interval=2)
+
+
+@pytest.mark.topospecific
+@pytest.mark.sanity
+@pytest.mark.syncgateway
+@pytest.mark.sgaccel
+@pytest.mark.parametrize("sg_conf", [
+    ("{}/missing_num_shards_di.json".format(SYNC_GATEWAY_CONFIGS)),
+])
+def test_missing_num_shards(params_from_base_test_setup, sg_conf):
+    """
+    1. Launch sg_accels missing the following property in the config.
+        "feed_params":{
+            "num_shards":64
+        }
+    2. Verify there are 64 shards
+    3. Verify they are distributed evenly across the nodes
+    """
+
+    cluster_conf = params_from_base_test_setup["cluster_config"]
+
+    log_info("Running 'test_missing_num_shards'")
+    log_info("cluster_conf: {}".format(cluster_conf))
+    log_info("sg_conf: {}".format(sg_conf))
+
+    cluster = Cluster(config=cluster_conf)
+    cluster.reset(sg_config_path=sg_conf)
+
+    # CBGT REST Admin API endpoint
+    admin_api = Admin(cluster.sg_accels[1])
+    cbgt_cfg = admin_api.get_cbgt_cfg()
+
+    # Verify that 64 shards are being used
+    assert len(cbgt_cfg["planPIndexes"]["planPIndexes"]) == 64
+
+    # Verify sharding is correct
+    cluster.validate_cbgt_pindex_distribution_retry(3)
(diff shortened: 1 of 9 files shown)


tleyden:
@sethrosetter is there any way to make this test more robust against changing default values?  For example, if we change the default val from 64, this will break and will be a little hard to figure out why.

testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
```diff
@@ -451,3 +451,40 @@ def test_take_all_sgaccels_down(params_from_base_test_setup, sg_conf):
...
+
+    # CBGT REST Admin API endpoint
+    admin_api = Admin(cluster.sg_accels[1])
+    cbgt_cfg = admin_api.get_cbgt_cfg()
+
+    # Verify that 64 shards are being used
+    assert len(cbgt_cfg["planPIndexes"]["planPIndexes"]) == 64
+
+    # Verify sharding is correct
+    cluster.validate_cbgt_pindex_distribution_retry(3)
```
tleyden: Can this be changed to `cluster.validate_cbgt_pindex_distribution_retry(num_running_sg_accels=3)`?

testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
```diff
@@ -451,3 +451,40 @@ def test_take_all_sgaccels_down(params_from_base_test_setup, sg_conf):
...
+
+    cluster = Cluster(config=cluster_conf)
+    cluster.reset(sg_config_path=sg_conf)
+
+    # CBGT REST Admin API endpoint
+    admin_api = Admin(cluster.sg_accels[1])
+    cbgt_cfg = admin_api.get_cbgt_cfg()
+
+    # Verify that 64 shards are being used
+    assert len(cbgt_cfg["planPIndexes"]["planPIndexes"]) == 64
```
tleyden: This is really cryptic .. and I know we probably have this in other places in the code.  What if we created a wrapper around the cbgt stuff that made it easier to understand, like:

```
cbgt_configuration = admin_api.get_cbgt_configuration()
assert cbgt_configuration.num_shards() == 64
```

and then in the `num_shards()` method it could have some documentation that explains what this is all about.

tleyden:
I guess what I'm suggesting is that this check might be enough `cluster.validate_cbgt_pindex_distribution_retry(num_running_sg_accels=3)`, rather than asserting that numshards == 64.

testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
sethrosetter: I like that. Will update

sethrosetter:
Are we expecting to change the default NUM_SHARDS?

tleyden:
> Are we expecting to change the default NUM_SHARDS?

No.  But it wouldn't shock me since there was a bit of uncertainty around choosing this value.

sethrosetter:
I created a CbgtConfig object per one of you suggestions and added a comment to the hard coded value. I would like to be explicit in that assertion as it would be an easy future fix. Let me know if you feel strongly about that.

testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
```diff
@@ -451,3 +451,41 @@ def test_take_all_sgaccels_down(params_from_base_test_setup, sg_conf):
...
+
+    log_info("Running 'test_missing_num_shards'")
+    log_info("cluster_conf: {}".format(cluster_conf))
+    log_info("sg_conf: {}".format(sg_conf))
+
+    cluster = Cluster(config=cluster_conf)
+    cluster.reset(sg_config_path=sg_conf)
+
+    # CBGT REST Admin API endpoint
+    admin_api = Admin(cluster.sg_accels[1])
```
tleyden: why is this `cluster.sg_accels[1]` as opposed to `cluster.sg_accels[0]`?  Also, how is it guaranteed that there are at least 2 sg accels?

tleyden: Instead of arbitrarily picking `cluster.sg_accels[1]`, what about calling `cluster.get_random_sg_accel()` to return one of the sg_accels?  If the test is ever run in a context where it only has a single sg accel, it will still work.

tleyden: Merging since this looks good, we can file a separate ticket if this seems like it's worth changing

PR #855, opened 2017-01-11
Fixes mono download path check (it is incorrect so it gets downloaded…

borrrden:
… every time even if it exists) and adds similar logic to MSFT

- [X] Ran `flake8`

- Check the correct directory for existing Mono LiteServ
- Perform similar logic for Windows LiteServ

### DIFF
--- a/keywords/LiteServNetMsft.py
+++ b/keywords/LiteServNetMsft.py
@@ -52,12 +52,6 @@ class LiteServNetMsft(LiteServBase):
         2. Extracts the package and removes the zip
         """
 
-        # Skip download if packages is already downloaded
-        expected_binary = "couchbase-lite-net-msft-{}-liteserv/net45/LiteServ.exe".format(self.version_build)
-        if os.path.isfile(expected_binary):
-            log_info("Package already downloaded: {}".format(expected_binary))
-            return
-
         version, build = version_and_build(self.version_build)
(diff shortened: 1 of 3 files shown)


keywords/LiteServNetMsft.py
```diff
@@ -52,6 +52,12 @@ def download(self):
         2. Extracts the package and removes the zip
         """
 
+        # Skip download if packages is already downloaded
+        expected_binary = "couchbase-lite-net-msft-{}-liteserv/net45/LiteServ.exe".format(self.version_build)
+        if os.path.isfile(expected_binary):
+            log_info("Package already downloaded: {}".format(expected_binary))
+            return
+
```
sethrosetter: I'm not sure this would work. Basically, it needs to check via the ansible script if the directory exists on the remote machine.

borrrden: Yuck, does that mean that it is forced to delete and download it every time?

sethrosetter: We can make this smarter, i bet - https://github.com/couchbaselabs/mobile-testkit/blob/master/libraries/provision/ansible/playbooks/download-liteserv-msft.yml

borrrden: I was just looking at that.  I think I have something

borrrden:
This should do the trick!  Tested on uber jenkins

sethrosetter:
@borrrden Looks good!! Thanks!

PR #926, opened 2017-02-08
Change to stop liteserv and pull logs in case DB delete fails

raghusarangapani:
- [x] Ran `flake8`

- If client.delete_databases fails, catch the exception
- stop liteserv and fetch logs.

### DIFF
--- a/testsuites/listener/shared/client_sg/conftest.py
+++ b/testsuites/listener/shared/client_sg/conftest.py
@@ -10,7 +10,6 @@ from keywords.MobileRestClient import MobileRestClient
 from keywords.ClusterKeywords import ClusterKeywords
 from keywords.SyncGateway import sync_gateway_config_path_for_mode
 from keywords.Logging import Logging
-from requests.models import HTTPError
 
 
 # Add custom arguments for executing tests in this directory
--- a/testsuites/listener/shared/client_sg/conftest.py
+++ b/testsuites/listener/shared/client_sg/conftest.py
@@ -124,6 +124,7 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
 
     try:
         client.delete_databases(ls_url)
+        liteserv.stop()
     except:
         liteserv.stop()
         logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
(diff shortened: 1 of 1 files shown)


testsuites/listener/shared/client_sg/conftest.py
```diff
@@ -119,10 +121,17 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
...
-        logging_helper = Logging()
-        logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
+    try:
+        client.delete_databases(ls_url)
+    except HTTPError as he:
+        liteserv.stop()
+        raise HTTPError(he)
+    except:
+        liteserv.stop()
+        raise HTTPError(sys.exc_info()[0])
```
sethrosetter: Can you leave a comment about when you think this will occur?

testsuites/listener/shared/client_sg/conftest.py
```diff
@@ -119,10 +121,17 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
...
+    finally:
+        # if the test failed pull logs
+        if request.node.rep_call.failed:
```
sethrosetter: If an exception in throw in this fixture, are you sure this will be true? One other option is to take this out of the finally and make sure to pull after to stoping the liteserv in the exception handler. I think that may be simpler

testsuites/listener/shared/client_sg/conftest.py
```diff
@@ -119,10 +120,11 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
...
     log_info("Tearing down test")
-    client.delete_databases(ls_url)
-    liteserv.stop()
+    logging_helper = Logging()
 
-    # if the test failed pull logs
-    if request.node.rep_call.failed:
-        logging_helper = Logging()
+    try:
+        client.delete_databases(ls_url)
```
sethrosetter: If this succeeds, I don't think LiteServ is going to stop.

raghusarangapani: Done

PR #927, opened 2017-02-08
Handle cc / di user doc differences

sethrosetter:
- [x] Ran `flake8`

- Fix tests to handle _user/ doc behavior differences
- Disable bucket shadow tests in di mode

### DIFF
--- a/testsuites/syncgateway/functional/tests/test_changes_backfill.py
+++ b/testsuites/syncgateway/functional/tests/test_changes_backfill.py
@@ -372,6 +372,10 @@ def test_backfill_channels_oneshot_limit_changes(params_from_base_test_setup, sg
     # Changes Req #1
     #################
 
+    # Expect a user doc in the changes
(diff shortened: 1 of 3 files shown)


sethrosetter:
Just so I'm clear `CHANNEL-TO-ROLE-REST` and `CHANNEL-TO-ROLE-SYNC` should not see any user docs in the backfill, correct?

adamcfraser:
@seth That's correct - the channel-to-role grant doesn't trigger a user doc update, so even a REST-based update shouldn't result in a user doc update.  We also don't have the concept of pseudo-role docs in the feed (i.e. we don't send a _role/rolename entry).

tleyden:
> Disable bucket shadow tests in di mode

@adamcfraser If we are doing this, shouldn't we detect this config combo altogether to prevent users running in this mode?

adamcfraser:
@tleyden Would be a nice enhancement but not a must-have for 1.4.

Issue #804, opened 2016-12-06
Figure out why sync_gateway config template parser blows up with todo config


sethrosetter:
Prerequisites:
- Follow the instructions on provisioning a cluster in the README.md
- Make sure to target vagrant boxes
- Get a local cluster of vagrant vms launched and provisioned with the default sync_gateway config. 
- export CLUSTER_CONFIG=resources/cluster_config/1sg_1cbs
- The sync_gateway configuration parsing may be a good place to start - https://github.com/couchbaselabs/mobile-testkit/blob/master/mobile_testkit_tests/test_install_sync_gateway.py#L52-L106

This is the problem

When provisioning  a cluster, the sync_gateway configs will we rendered fromt the below templates. We are parsing these to get bucket names and in the case of the below configs, we we need to figure out why. 

To execute the provisioning task
```
python libraries/provision/provision_cluster.py --server-version=4.5.1 --sync-gateway-version=1.4.0-12 --sync-gateway-config=/Users/sethrosetter/Code/mobile-testkit/resources/sync_gateway_configs/todo_cc.json
```

To check in code
- Create a new branch `git checkout -b fix/template-parsing`

Closing criteria
- Needs to avoid crash when provisioning and successfully deploy and launch the sync_gateway with the below config
- Needs unit test in mobile_testkit_tests/

sethrosetter:
Config 1

```
{
  "interface":":4984",
  "log": ["HTTP", "Auth"],
  "databases": {
    "todo": {
      "server":"http://{{ couchbase_server_primary_node }}:8091",
      "bucket": "data-bucket",
      "users": {
        "user1": {"password": "pass", "admin_channels": ["user1"]},
        "user2": {"password": "pass", "admin_channels": ["user2"]},
        "mod": {"password": "pass", "admin_roles": ["moderator"]},
        "admin": {"password": "pass", "admin_roles": ["admin"]}
      },
      "roles": {
        "moderator": {},
        "admin": {}
      },
      "sync": `
function(doc, oldDoc){
  /* Type validation */
  if (isCreate()) {
    // Don't allow creating a document without a type.
    validateNotEmpty("type", doc.type);
  } else if (isUpdate()) {
    // Don't allow changing the type of any document.
    validateReadOnly("type", doc.type, oldDoc.type);
  }

  if (getType() == "moderator") {
    /* Control Write Access */
    // Only allow admins to add/remove moderators.
    requireRole("admin");

    /* Validate */
[...]

PR #945, opened 2017-02-11
Fix the sync gateway config parser

raghusarangapani:
#### Fixes #.
https://github.com/couchbaselabs/mobile-testkit/issues/804

- [x] Ran `flake8`

- Ignore the sync function in the sync gateway configs to get the bucket names
- Added a unit test for get_buckets_from_sync_gateway_config method - 100% pass

### DIFF
--- a/libraries/provision/install_sync_gateway.py
+++ b/libraries/provision/install_sync_gateway.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import re
 
 from keywords.couchbaseserver import CouchbaseServer
 from keywords.ClusterKeywords import ClusterKeywords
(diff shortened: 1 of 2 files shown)


tleyden:
> Added a unit test for get_buckets_from_sync_gateway_config method - 100% pass

Neat!  We should really be running unit tests on each PR though.

We have drone.couchbase.io setup and running, and it's testing Sync Gateway on every commit.

For a python example, checkout [capitalone/cloud-custodian/.../.drone.yml](https://github.com/capitalone/cloud-custodian/blob/master/.drone.yml)

Anyone on the couchbase team should already have access to drone.couchbase.io, login with github.

PR #947, opened 2017-02-13
Catch HTTP Error specifically

raghusarangapani:
#### Fixes #.
LiteServ was being stopped upon an exception when it was already stopped. Now we only try to stop it if the exception thrown is an HTTPError.

- [ ] Ran `flake8`

LiteServ was being stopped upon an exception when it was already stopped. Now we only try to stop it if the exception thrown is an HTTPError.

### DIFF
--- a/testsuites/listener/shared/client_sg/conftest.py
+++ b/testsuites/listener/shared/client_sg/conftest.py
@@ -129,7 +129,7 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
     except HTTPError as h:
         liteserv.stop()
         logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
-        raise Exception(h)
+        raise HTTPError(h)
     except:
         logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
         raise Exception("Something went wrong trying to delete the database", sys.exc_info()[0])
--- a/testsuites/listener/shared/client_sg/conftest.py
+++ b/testsuites/listener/shared/client_sg/conftest.py
@@ -127,9 +127,11 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
         client.delete_databases(ls_url)
         liteserv.stop()
     except HTTPError as h:
+        # Stop liteserv, save the logs and Rethrow the exception caught
         liteserv.stop()
         logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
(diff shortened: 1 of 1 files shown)


testsuites/listener/shared/client_sg/conftest.py
```diff
@@ -125,7 +126,10 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
     try:
         client.delete_databases(ls_url)
         liteserv.stop()
-    except:
+    except HTTPError as h:
         liteserv.stop()
         logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
-        raise Exception(sys.exc_info()[0])
+        raise Exception(h)
```
sethrosetter: Please raise an HTTPError here.

raghusarangapani: Done

testsuites/listener/shared/client_sg/conftest.py
```diff
@@ -125,7 +126,10 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
     try:
         client.delete_databases(ls_url)
         liteserv.stop()
-    except:
+    except HTTPError as h:
         liteserv.stop()
         logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
-        raise Exception(sys.exc_info()[0])
+        raise HTTPError(h)
```
sethrosetter: Is it ok to wrap an exception in an exception? Why not just print a message in the exception that you raise?

testsuites/listener/shared/client_sg/conftest.py
```diff
@@ -125,7 +126,10 @@ def setup_client_syncgateway_test(request, setup_client_syncgateway_suite):
...
+    except:
+        logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
+        raise Exception("Something went wrong trying to delete the database", sys.exc_info()[0])
```
sethrosetter: What will sys.exc_info()[0] print?

testsuites/listener/shared/client_sg/conftest.py
raghusarangapani: Done, just Rethrow the exception caught

testsuites/listener/shared/client_sg/conftest.py
raghusarangapani: Done, save the logs and Rethrow the exception caught

Issue #979, opened 2017-02-20
Fix rollback test to handle clusters with multiple Couchbase Servers


sethrosetter:
The vbucket file may not be on the server we are targeting. 

- Reenable when fixed - https://github.com/couchbaselabs/mobile-testkit/blob/master/testsuites/syncgateway/functional/tests/test_rollback.py#L17-L24

sethrosetter:
Sync with @tleyden or @adamcfraser before fixing this. 

The reason it works with one node server cluster and not 3. 
Testkit is not multinode aware when trying to delete the vbucket. 

The fix:
testkit needs to know how to resolve an Server node endpoint given a vBucket.

sridevi-15:
Sync up with @adamcfraser for priority

sridevi-15:
@adamcfraser , Could you let me know the expected behavior after multiple url support? 
Test still fails with multiple url support as test is  expecting sync-gateway fail to connect to server to get changes when server is restarted. It should retry atleast once to get changes through sync-gateway .


This is the code : 
https://github.com/couchbaselabs/mobile-testkit/blob/master/testsuites/syncgateway/functional/tests/test_rollback.py#L119-L137

It fails at Line 137.

tleyden:
From discussion w / @sridevi-15 

- This comment from Seth https://github.com/couchbaselabs/mobile-testkit/issues/979#issuecomment-317901639 mentions that the test will need to be changed to delete the vbucket from the correct couchbase server node in order for the test to work with multiple couchbase server nodes.
- In addition to deleting the vbucket on the primary node, the vbucket may need to be deleted on the replica copy as well in order to simulate rollback behavior
- The `find -delete` command could be failing silently, needs more investigation to see what the process exit code when `find -delete` is called on non-existent file

tleyden:
@sridevi-15 does this test pass or fail with a single couchbase server node?

sridevi-15:
sg-config : 
```{
    "interface":":4984",
    "adminInterface": "0.0.0.0:4985",
    "maxIncomingConnections": 0,
    "maxCouchbaseConnections": 16,
    "maxFileDescriptors": 90000,
    "slowServerCallWarningThreshold": 500,
    "compressResponses": false,
    "log": ["CRUD+", "Cache+", "HTTP+", "Changes+", "DCP"],
    "cluster_config": {
        "server":"http://192.168.33.10,192.168.33.11,192.168.33.12:8091",
        "data_dir":".",
[...]

PR #980, opened 2017-02-20
Skip roll back test due to test errors with multiple server cluster

sethrosetter:
- [x] Ran `flake8`

- We need to disable this because it is causing issues when running in CI -  https://github.com/couchbaselabs/mobile-testkit/issues/979


raghusarangapani:
Looks good

PR #996, opened 2017-03-01
Add 1.4.0 to the LiteServ path check

raghusarangapani:
- [ ] Ran `flake8`

1.4.0 does not have android, net45, ios folders. This change adds a check to select the right path for LiteServ.exe

### DIFF
--- a/keywords/LiteServNetMono.py
+++ b/keywords/LiteServNetMono.py
@@ -14,6 +14,7 @@ from keywords.constants import REGISTERED_CLIENT_DBS
 from keywords.exceptions import LiteServError
 from keywords.utils import version_and_build
 from keywords.utils import log_info
+from keywords.utils import hasDotNet4Dot5
 
 
 class LiteServNetMono(LiteServBase):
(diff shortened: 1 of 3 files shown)


keywords/LiteServNetMono.py
```diff
@@ -26,7 +26,11 @@ def download(self):
         """
 
         # Skip download if packages is already downloaded
-        expected_binary = "{}/couchbase-lite-net-mono-{}-liteserv/net45/LiteServ.exe".format(BINARY_DIR, self.version_build)
+        if self.version_build.startswith("1.2") or self.version_build.startswith("1.3") or self.version_build.startswith("1.4.0"):
```
tleyden: any reason not to use `self.version_build.startswith("1.4")` to be consistent with other comparisons?

raghusarangapani: For 1.4, the path is net45/LiteServ.exe, for 1.4.0, there is no net45

tleyden: That seems a bit confusing.

What about extracting a method like:

```
# Check if this version has net45
def hasDotNet4Dot5():
    version_prefixes = [
         "1.2", 
         "1.3", 
         "1.4.0",  # For 1.4, the path is net45/LiteServ.exe, for 1.4.0, there is no net45
         ...
]
   .. 
   return True or False
   
```

and then maybe call that function in the various places rather than duplicating the conditional logic

raghusarangapani:
Done. Launched a job with my branch, it is working fine

PR #1083, opened 2017-03-29
Feature/enable running races for sg tests

sethrosetter:
#### Updating log collection on every run.

- [x] Ran `flake8`

- Allow sync gateway functional test to run with `--race` flag. When this is enabled with a commit as the version, Sync Gateway will be built with -race. 

- Allow sync gateway functional test to run with `--collect-logs` flag. When this is enabled, every test will pull logs regardless of pass / fail.


tleyden:
@raghusarangapani can you review this?  (assuming it's ready .. @sethrosetter can you assign @raghusarangapani when it's ready?)

raghusarangapani:
Looks good to me

PR #1111, opened 2017-04-11
Add rbac exceptions

raghusarangapani:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Adds RBACUserCreationException and RBACUserDeletionException
- Throws these exceptions for RBAC operations

### DIFF
--- a/keywords/couchbaseserver.py
+++ b/keywords/couchbaseserver.py
@@ -15,6 +15,8 @@ from keywords.remoteexecutor import RemoteExecutor
 from keywords.exceptions import CBServerError
 from keywords.exceptions import ProvisioningError
 from keywords.exceptions import TimeoutError
+from keywords.exceptions import RBACUserCreationError
+from keywords.exceptions import RBACUserDeletionError
 from keywords.utils import log_r
 from keywords.utils import log_info
 from keywords.utils import log_debug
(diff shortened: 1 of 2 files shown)


keywords/couchbaseserver.py
```diff
@@ -81,7 +83,7 @@ def create_internal_rbac_bucket_user(url, bucketname):
         resp.raise_for_status()
     except HTTPError as h:
         log_info("resp code: {}; error: {}".format(resp, h))
-        raise
+        raise RBACUserCreationError(h)
```
sethrosetter: A simple string would be fine as a parameter, i.e.

`raise RBACUserCreationError('Could not create RBAC user!')`

keywords/couchbaseserver.py
```diff
@@ -101,7 +103,7 @@ def delete_internal_rbac_bucket_user(url, bucketname):
         resp.raise_for_status()
     except HTTPError as h:
         log_info("resp code: {}; error: {}".format(resp, h))
-        raise
+        raise RBACUserDeletionError(h)
```
sethrosetter: A simple string would be fine as a parameter, i.e.

`raise RBACUserCreationError('Could not delete RBAC user!')`

keywords/couchbaseserver.py
sethrosetter: Nevermind

keywords/couchbaseserver.py
sethrosetter: Nevermind

PR #1157, opened 2017-04-27
Add method to wait until a db is in the online state

tleyden:
I wrote and tested this while debugging https://github.com/couchbase/sync_gateway/issues/2423#issuecomment-297576701 and it works.  Seems like a useful addition

#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Add method to wait until a db is in the online state

### DIFF
--- a/libraries/testkit/user.py
+++ b/libraries/testkit/user.py
@@ -143,6 +143,20 @@ class User:
 
         return doc_id
 
+    def wait_until_db_online(self):
+        db_url = self.target.url + "/" + self.db + "/"
+        while True:
+            log.debug("DB [{}] waiting until online".format(self.db))
+            resp = self._session.get(db_url, timeout=settings.HTTP_REQ_TIMEOUT)
+            resp.raise_for_status()
+            resp_json = resp.json()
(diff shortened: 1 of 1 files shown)


libraries/testkit/user.py
```diff
@@ -143,6 +143,20 @@ def add_doc(self, doc_id=None, content=None, retries=False):
 
         return doc_id
 
+    def wait_until_db_online(self):
+        db_url = self.target.url + "/" + self.db + "/"
+        while True:
+            log.debug("DB [{}] waiting until online".format(self.db))
+            resp = self._session.get(db_url, timeout=settings.HTTP_REQ_TIMEOUT)
+            resp.raise_for_status()
```
raghusarangapani: A try catch might be needed here. This might throw an HTTP error

tleyden: Good point

It should probably only try a limited amount of times before giving up too

sethrosetter:
We dont have a need for this at the moment. Thanks though!

PR #1177, opened 2017-05-05
Fix the net45 folder dependency

### LINKED ISSUE #1178: Fix the net45 folder dependency for .net liteserv
Some .net builds have a net45 folder for the LiteServ and some don't. We currently maintain a list of versions that has the net45 folder. This is not scalable. We need a way to dynamically retry with and without net45 folder.

raghusarangapani:
#### Fixes #1178 .

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Some LiteServ version for .net have a `net45` folder and some don't. The current implementation breaks the runs every time there is a new build 
- This change adds a retry mechanism, first without `net45` in the path and then with `net45` in the path


raghusarangapani:
The change has been tested:
http://uberjenkins.sc.couchbase.com:8080/job/net-mono-listener-client-sync-gateway/392/
http://uberjenkins.sc.couchbase.com/job/net-windows-listener-client-sync-gateway/398/

PR #1187, opened 2017-05-11
Skip offline tests in DI mode

tleyden:
see https://github.com/couchbase/sync_gateway/issues/2423#issuecomment-300841425

#### Fixes https://github.com/couchbase/sync_gateway/issues/2423#issuecomment-300841425.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Skip offline tests in DI mode
- I moved the `rest_scan` function to the bottom of the file since it was confusing to see it at the top... when I was expecting to see actual test cases.

### DIFF
--- a/testsuites/syncgateway/functional/tests/test_db_online_offline.py
+++ b/testsuites/syncgateway/functional/tests/test_db_online_offline.py
@@ -20,6 +20,7 @@ from keywords.SyncGateway import sync_gateway_config_path_for_mode
 
 NUM_ENDPOINTS = 13
 
+
 # Scenario 1
 @pytest.mark.sanity
 @pytest.mark.syncgateway
--- a/testsuites/syncgateway/functional/tests/test_db_online_offline.py
+++ b/testsuites/syncgateway/functional/tests/test_db_online_offline.py
@@ -20,6 +20,721 @@ from keywords.SyncGateway import sync_gateway_config_path_for_mode
 
 NUM_ENDPOINTS = 13
 
+# Scenario 1
+@pytest.mark.sanity
+@pytest.mark.syncgateway
+@pytest.mark.onlineoffline
+@pytest.mark.basicauth
+@pytest.mark.role
+@pytest.mark.channel
+@pytest.mark.bulkops
+@pytest.mark.changes
+@pytest.mark.parametrize("sg_conf_name, num_docs", [
+    ("bucket_online_offline/bucket_online_offline_default", 100)
+])
+def test_online_default_rest(params_from_base_test_setup, sg_conf_name, num_docs):
+
+    cluster_conf = params_from_base_test_setup["cluster_config"]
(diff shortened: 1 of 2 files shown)


testsuites/syncgateway/functional/tests/test_db_online_offline.py
```diff
@@ -873,6 +737,178 @@ def test_multiple_dbs_unique_buckets_lose_tap(params_from_base_test_setup, sg_co
             assert error_tuple[1] == 503
 
 
+def rest_scan(sync_gateway, db, online, num_docs, user_name, channels):
```
sethrosetter: What was the reason for moving this? Has anything changed in here?

tleyden: Nope, nothing has changed.  I moved the `rest_scan` function to the bottom of the file since it was confusing to see it at the top... when I was expecting to see actual test cases.

tleyden:
From unclebob's [Clean Code](https://books.google.com/books/about/Clean_Code.html?id=_i6bDeoCQzsC):

Reading Code from Top to Bottom: The Stepdown Rule We want the code to read like a top-down narrative. 5 We want every function to be followed by those at the next level of abstraction so that we can read the program, descending one level of abstraction at a time as we read down the list of functions. I call this The Step-down Rule. 

To say this differently, we want to be able to read the program as though it were a set of TO paragraphs, each of which is describing the current level of abstraction and referencing subsequent TO paragraphs at the next level down. 

* **To** include the setups and teardowns, we include setups, then we include the test page content, and then we include the teardowns. 
* **To** include the setups, we include the suite setup if this is a suite, then we include the regular setup. To include the suite setup, we search the parent hierarchy for the “SuiteSetUp” page and add an include statement with the path of that page. 
* **To** search the parent… 

It turns out to be very difficult for programmers to learn to follow this rule and write functions that stay at a single level of abstraction. But learning this trick is also very important. It is the key to keeping functions short and making sure they do “one thing.” Making the code read like a top-down set of TO paragraphs is an effective technique for keeping the abstraction level consistent.

tleyden:
In [cen7-sync-gateway-functional-tests-base-di/508](http://uberjenkins.sc.couchbase.com/job/cen7-sync-gateway-functional-tests-base-di/508/console), which has this commit, the offline tests are being skipped in di mode:

```
testsuites/syncgateway/functional/tests/test_db_online_offline.py::test_db_offline_tap_loss_sanity[bucket_online_offline/bucket_online_offline_default_dcp-100] SKIPPED
12:45:09 testsuites/syncgateway/functional/tests/test_db_online_offline.py::test_db_offline_tap_loss_sanity[bucket_online_offline/bucket_online_offline_default-100] SKIPPED
```

sethrosetter:
Looks good if you file a documentation ticket for offline not supported in Accel.

PR #1228, opened 2017-06-06
Changes for iOS simulator runs

### LINKED ISSUE #1227: Add ios simulator support for p2p tests
The p2p tests currently cannot be run on ios device or simulator. We should add support to run p2p tests on ios simulator as a first step.

raghusarangapani:
#### Fixes #1227.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Run p2p tests using ios simulator

### DIFF
--- a/keywords/LiteServBase.py
+++ b/keywords/LiteServBase.py
@@ -52,8 +52,11 @@ class LiteServBase(object):
         log_r(resp)
         raise LiteServError("There should be no service running on the port")
 
-    def _wait_until_reachable(self):
-        url = "http://{}:{}".format(self.host, self.port)
+    def _wait_until_reachable(self, port=None):
+        if not port:
+            port = self.port
+
+        url = "http://{}:{}".format(self.host, port)
         count = 0
         while count < MAX_RETRIES:
             try:
--- a/keywords/LiteServiOS.py
+++ b/keywords/LiteServiOS.py
@@ -300,6 +300,8 @@ class LiteServiOS(LiteServBase):
         # Have to separately copy the simulator logs
         if self.logfile_name and self.device_id:
             home = os.environ['HOME']
-            copyfile("{}/Library/Logs/CoreSimulator/{}/system.log".format(home, self.device_id), self.logfile_name)
+            ios_log_file = "{}/Library/Logs/CoreSimulator/{}/system.log".format(home, self.device_id)
+            copyfile(ios_log_file, self.logfile_name)
+            open(ios_log_file, 'w').close()
 
         self._verify_not_running()
--- a/keywords/LiteServiOS.py
+++ b/keywords/LiteServiOS.py
@@ -4,6 +4,7 @@ import os
 import re
 import time
 from zipfile import ZipFile
+from shutil import copyfile
 import requests
 
 from keywords.LiteServBase import LiteServBase
@@ -18,6 +19,12 @@ from keywords.constants import CLIENT_REQUEST_TIMEOUT
 
 
 class LiteServiOS(LiteServBase):
+
+    def __init__(self, version_build, host, port, storage_engine):
+        super(LiteServiOS, self).__init__(version_build, host, port, storage_engine)
+        self.liteserv_admin_url = "http://{}:59850".format(self.host)
+        self.logfile_name = None
+        self.device_id = None
 
     def download(self):
         """
@@ -137,6 +144,17 @@ class LiteServiOS(LiteServBase):
                 time.sleep(1)
                 continue
 
+        # Get the device ID
+        list_output = subprocess.Popen(["xcrun", "simctl", "list"], stdout=subprocess.PIPE)
+        output = subprocess.check_output(('grep', 'Booted'), stdin=list_output.stdout)
+
+        for line in output.splitlines():
+            if "Phone" in line:
+                self.device_id = re.sub(' +', ' ', line).strip()
+                self.device_id = self.device_id.split(" ")[4]
+                self.device_id = self.device_id.strip('(')
+                self.device_id = self.device_id.strip(')')
+
     def remove_device(self):
         """
         Remove the iOS app from the connected device
@@ -167,27 +185,15 @@ class LiteServiOS(LiteServBase):
 
         self.stop()
 
-        # Get the device ID
-        device_id = None
-        list_output = subprocess.Popen(["xcrun", "simctl", "list"], stdout=subprocess.PIPE)
-        output = subprocess.check_output(('grep', 'Booted'), stdin=list_output.stdout)
-
-        for line in output.splitlines():
-            if "Phone" in line:
-                device_id = re.sub(' +', ' ', line).strip()
-                device_id = device_id.split(" ")[4]
-                device_id = device_id.strip('(')
-                device_id = device_id.strip(')')
-
         # Stop the simulator
-        log_info("device_id: {}".format(device_id))
+        log_info("device_id: {}".format(self.device_id))
         output = subprocess.check_output([
             "killall", "Simulator"
         ])
 
         # Erase the simulator
         output = subprocess.check_output([
-            "xcrun", "simctl", "erase", device_id
+            "xcrun", "simctl", "erase", self.device_id
         ])
 
         if bundle_id in output:
@@ -205,6 +211,7 @@ class LiteServiOS(LiteServBase):
         data = {}
         encryption_enabled = False
         device = "iPhone-7-Plus"
+        self.logfile_name = logfile_name
 
         package_name = "LiteServ-iOS.app"
         app_dir = "LiteServ-iOS"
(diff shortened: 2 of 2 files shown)


keywords/LiteServiOS.py
```diff
@@ -51,13 +87,61 @@ def install(self):
...
+        package_name = "LiteServ-iOS.app"
+        app_dir = "LiteServ-iOS"
+
+        if self.storage_engine == "SQLCipher":
+            package_name = "LiteServ-iOS-SQLCipher.app"
+            app_dir = "LiteServ-iOS-SQLCipher"
+
+        app_path = "{}/{}/{}".format(BINARY_DIR, app_dir, package_name)
+        output = subprocess.check_output([
+            "ios-sim", "--devicetypeid", device, "start"
```
sethrosetter: What happens if simulator is started already?

keywords/LiteServiOS.py
```diff
@@ -51,13 +87,61 @@ def install(self):
...
+
+        log_info(output)
+
+        # Wait for the device to boot up
+        start = time.time()
+        while True:
+            if time.time() - start > CLIENT_REQUEST_TIMEOUT:
+                raise LiteServError("iPhone Simulator failed to start")
+
+            output = subprocess.Popen([
```
sethrosetter: Can you add a comment to highlight what you are doing here?

keywords/LiteServiOS.py
```diff
@@ -51,13 +87,61 @@ def install(self):
...
+            status = output.split(" ")[2]
+            if status == "A":
+                log_info("iPhone Simulator seems to have booted up")
+                break
+            else:
+                log_info("Waiting for the iPhone Simulator to boot up")
+                time.sleep(1)
+                continue
+
+    def remove_device(self):
```
sethrosetter: Why are you touching remove device if we are only working with simulator?

keywords/LiteServiOS.py
```diff
@@ -70,6 +154,44 @@ def remove(self):
...
+        if self.storage_engine == "SQLCipher":
+            bundle_id = "com.couchbase.LiteServ-iOS-SQLCipher"
+
+        log_info("Removing LiteServ")
+
+        self.stop()
+
+        # Get the device ID
+        device_id = None
+        list_output = subprocess.Popen(["xcrun", "simctl", "list"], stdout=subprocess.PIPE)
```
sethrosetter: What if multiple simulators are running for multiple test runs?

keywords/LiteServBase.py
```diff
@@ -52,8 +52,11 @@ def _verify_not_running(self):
         log_r(resp)
         raise LiteServError("There should be no service running on the port")
 
-    def _wait_until_reachable(self):
-        url = "http://{}:{}".format(self.host, self.port)
+    def _wait_until_reachable(self, port=None):
+        if not port:
```
raghusarangapani: The LiteServ iOS starts off with the default admin port 59850 and public port 49850, which is not the same as the --liteserv-port we specify on the command line. Switching to the new port specified on the command line happens through a REST API call to /start. The _wait_until_reachable takes the port specified on the command line to check for the status as soon as the LiteServ is launched. This will obviously time out as the /start REST API call happens much later, when the tests actually run. This change enables checking the status with any port. The default behavior(self.port) is not changed.

keywords/LiteServiOS.py
raghusarangapani: No error is raised. It will go ahead and install the .app

keywords/LiteServiOS.py
raghusarangapani: Done

keywords/LiteServiOS.py
raghusarangapani: I was not able to successfully launch multiple simulators, though some articles on the internet mentioned it is possible.

keywords/LiteServiOS.py
raghusarangapani: Removed the 2 line change

keywords/LiteServiOS.py
```diff
@@ -79,25 +204,56 @@ def start(self, logfile_name):
...
+            db_flags = []
+            for db_name in REGISTERED_CLIENT_DBS:
+                db_flags.append("{}:pass".format(db_name))
+            db_flags = ",".join(db_flags)
+
+            log_info("Running with db_flags: {}".format(db_flags))
+            data["dbpasswords"] = db_flags
+
+        self._wait_until_reachable(port=59850)
+        liteserv_admin_url = "http://{}:59850".format(self.host)
```
sethrosetter: Since this is used multiple times, can you store as a class property?

keywords/LiteServiOS.py
```diff
@@ -131,8 +287,20 @@ def stop(self):
 
         liteserv_admin_url = "http://{}:59850".format(self.host)
         log_info("Stopping LiteServ: {}".format(liteserv_admin_url))
-        resp = self.session.put("{}/stop".format(liteserv_admin_url))
-        log_r(resp)
-        resp.raise_for_status()
+
+        count = 0
+        while count < MAX_RETRIES:
```
sethrosetter: Why are you retrying here? `liteserv.stop` should only be called if a LiteServ is running.

keywords/LiteServiOS.py
raghusarangapani: I thought that this was specific to iOS, I don't see such a usage anywhere in LiteServAndroid.py.

keywords/LiteServiOS.py
raghusarangapani: Good catch, this was left over code from my experiments.

keywords/LiteServiOS.py
sethrosetter: it is iOS specific. Store it as a `LiteServiOS.py` property

keywords/LiteServiOS.py
```diff
@@ -137,6 +144,17 @@ def install(self):
                 time.sleep(1)
                 continue
 
+        # Get the device ID
+        list_output = subprocess.Popen(["xcrun", "simctl", "list"], stdout=subprocess.PIPE)
```
sethrosetter: why not use `subprocess.check_ouput()` here?

keywords/LiteServiOS.py
```diff
@@ -137,6 +144,17 @@ def install(self):
...
+        output = subprocess.check_output(('grep', 'Booted'), stdin=list_output.stdout)
+
+        for line in output.splitlines():
+            if "Phone" in line:
+                self.device_id = re.sub(' +', ' ', line).strip()
+                self.device_id = self.device_id.split(" ")[4]
+                self.device_id = self.device_id.strip('(')
+                self.device_id = self.device_id.strip(')')
```
sethrosetter: Should we raise an exception is the device id is still `None` at this point?

keywords/LiteServiOS.py
```diff
@@ -282,12 +290,18 @@ def stop(self):
...
         log_r(resp)
         resp.raise_for_status()
 
+        # Using --exit in ios-sim means, --log has no effect
+        # Have to separately copy the simulator logs
+        if self.logfile_name and self.device_id:
+            home = os.environ['HOME']
+            ios_log_file = "{}/Library/Logs/CoreSimulator/{}/system.log".format(home, self.device_id)
+            copyfile(ios_log_file, self.logfile_name)
+            open(ios_log_file, 'w').close()
```
sethrosetter: Why is this needed?

keywords/LiteServiOS.py
raghusarangapani: I am using check_output in the next line to grep Booted. From my research, this is equivalent to bash's pipe feature.

keywords/LiteServiOS.py
raghusarangapani: Done

keywords/LiteServiOS.py
raghusarangapani: Are the comments insufficient?

sethrosetter: Yes. Looks good.

keywords/LiteServiOS.py
sethrosetter: Thanks

keywords/LiteServiOS.py
```diff
@@ -1,39 +1,80 @@
...
+from keywords.constants import REGISTERED_CLIENT_DBS
+from keywords.constants import CLIENT_REQUEST_TIMEOUT
 
 
 class LiteServiOS(LiteServBase):
 
     def __init__(self, version_build, host, port, storage_engine):
-
-        # Initialize baseclass properies
         super(LiteServiOS, self).__init__(version_build, host, port, storage_engine)
```
sethrosetter: add a check for `forestdb` and `forestdb+encryption`. If the user tries to use these with iOS, raise an exception.

raghusarangapani: Done

PR #1230, opened 2017-06-07
CRUD webhook tests with and without filtering (docmeta / xattrs)

### LINKED ISSUE #1206: XATTRs: Test webhook triggers on import
When a document gets written via the SDK, then imported by Sync Gateway, it should trigger any webhook events configured in Sync Gateway.  The webhook should be triggered on import (not on the initial SDK write), and apply the usual webhook filtering.

I don't think this is covered by the current test suite - I expect the webhook tests are being run in xattr mode, but I don't think there are any tests that verify webhook events for SDK writes.

sethrosetter:
#### Fixes #1206 .

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Cleanup of webhook tests / configs
- Add test for full CRUD webhooks (docmeta / xattrs) with a filter and no filter

### DIFF
--- a/libraries/testkit/web_server.py
+++ b/libraries/testkit/web_server.py
@@ -1,11 +1,9 @@
-from BaseHTTPServer import BaseHTTPRequestHandler
-import time
 import json
-from BaseHTTPServer import HTTPServer
 import threading
-from libraries.testkit import settings
-import logging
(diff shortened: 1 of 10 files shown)


raghusarangapani:
Looks ok

PR #1245, opened 2017-06-14
Adding SG LB flag

raghusarangapani:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Adding a flag for SG LB in the entrypoint.sh


docker/mobile-testkit/entrypoint.sh
```diff
@@ -7,8 +7,9 @@
 # $6 = Sync Gateway version
```
sethrosetter: Can you add an example for each param if there is not one?

PR #1259, opened 2017-06-20
Feature/use docker connection

sethrosetter:
#### Fixes.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Use ansible docker connection rather than ssh
- Update the timeout to get all changes in accel test
- Bump the number of connections allowed in the connection pool to clean up warnings when using small default size


mobile_testkit_tests/test_generate_clusters_from_pool.py
```diff
@@ -40,7 +40,7 @@ def test_generate_clusters_from_pool_ips(cleanup_folder, cluster_conf, num_couch
     cluster_config_dir = test_data + "/cluster_configs/"
 
     # Run tests with mock_pool_ips.json for backward compatibility
-    generate_clusters_from_pool(ips_pool_file)
+    generate_clusters_from_pool(ips_pool_file, False)
```
raghusarangapani: Don't we want an option to specify use_docker=True here?

mobile_testkit_tests/test_generate_clusters_from_pool.py
```diff
@@ -88,7 +88,7 @@ def test_generate_clusters_from_pool_ip_to_node(cleanup_folder, cluster_conf, nu
     cluster_config_dir = test_data + "/cluster_configs/"
 
     # Run tests with mock_pool_ips.json for backward compatibility
-    generate_clusters_from_pool(ip_to_node_type_pool_file)
+    generate_clusters_from_pool(ip_to_node_type_pool_file, False)
```
raghusarangapani: Don't we want an option to specify use_docker=True here?

testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
```diff
@@ -136,7 +136,7 @@ def test_dcp_reshard_sync_gateway_comes_up(params_from_base_test_setup, sg_conf)
             log_info("{} Completed:".format(tag))
 
     # TODO better way to do this
-    time.sleep(60)
+    time.sleep(120)
```
raghusarangapani: Is this related to the docker change?

resources/sync_gateway_configs/sync_gateway_default_functional_tests_di.json
```diff
@@ -23,14 +23,6 @@
             "bucket":"data-bucket",
             "username":"data-bucket",
             "password": "password",
```
raghusarangapani: Is this related to the docker change?

libraries/testkit/user.py
```diff
@@ -27,6 +27,8 @@ def __init__(self, target, db, name, password, channels):
         self.target = target
 
         self._session = requests.Session()
+        adapter = requests.adapters.HTTPAdapter(pool_connections=100, pool_maxsize=100)
```
raghusarangapani: Do we want to set this to a higher number than 100?

docker/mobile-testkit/Dockerfile
```diff
@@ -45,9 +45,6 @@ RUN cp ansible.cfg.example ansible.cfg
 # Set the correct default user
 RUN sed -i 's/remote_user = vagrant/remote_user = root/' ansible.cfg
 
-# Create .ssh directory for key installation
-RUN mkdir -p /root/.ssh
-
 # set python env
 ENV PYTHONPATH=/opt/mobile-testkit/
 ENV ANSIBLE_CONFIG=/opt/ansible.cfg/
```
raghusarangapani: The trailing / is probably not needed since it is a config file

libraries/testkit/user.py
sethrosetter: This is the fix for all of those Connection Pool errors you see from time to time. This is sufficient for the scale of those tests. Also, a good read is https://laike9m.com/blog/requests-secret-pool_connections-and-pool_maxsize,89/. I recommend checking it out.

mobile_testkit_tests/test_generate_clusters_from_pool.py
sethrosetter: by default the tests are not testing docker so false will result in the same behavior as before.

testsuites/syncgateway/functional/topology_specific_tests/multiple_accels/test_multiple_accels.py
sethrosetter: No this is related to an intermittent test failure where changes are still processing from time to time.

docker/mobile-testkit/Dockerfile
sethrosetter: oh good catch!

raghusarangapani:
The changes look ok. Has this change been tested locally or on jenkins? If the tests are passing, I'll merge.

PR #1306, opened 2017-07-12
Update Testkit.Net

sethrosetter:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Update 2.0 lite package
- Use .NET Core 2.0 for base image to work around replication issue
- Pass in number of docs for client to create in longevity simulation
- Clean up non implemented HttpServer to keep repo clean. A reference to the partial implementation is there for when we need to revisit


apps/testkit.net/Testkit.Net.Tests/Longevity.cs
```diff
@@ -42,79 +44,79 @@ public Longevity(string syncGatewayUrl, double scenarioRuntimeMinutes)
...
-            for (int i = 0; i < numDocs; i++)
-            {
-                var doc = _db.GetDocument($"doc_{i}");
-                _db.Delete(doc);
-            }
-
-            stopWatch.Stop();
-            _replicator.Stop();
-            _db.Delete();
+            var numDocs = _maxDocs;
```
sridevi-15: Is it necessary to create another variable numDocs, can we use _maxDocs directly here

sethrosetter: Fixed

PR #1318, opened 2017-07-18
Add log parsing to system test

raghusarangapani:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Change the log scanning logic from simple if "string" in to take a regex.
- Add log scanning variables to system test.

All existing tests in test_scan_logs.py passed. I also separately verified the scanning logic.

### DIFF
--- a/testsuites/syncgateway/system/conftest.py
+++ b/testsuites/syncgateway/system/conftest.py
@@ -209,7 +209,8 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
     )
 
     if status != 0:
-        raise LogScanningError("Failed to check for errors in logs")
+        logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
+        raise LogScanningError("Out of memory errors found in the logs")
 
     sg_error_log = ['panic', 'data race', 'SIGSEGV', 'nil pointer dereference']
     sg_log_file = "/home/sync_gateway/logs/sync_gateway_error.log"
@@ -226,4 +227,5 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
         )
 
         if status != 0:
-            raise LogScanningError("Failed to check for errors in logs")
+            logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=test_name)
+            raise LogScanningError("{} errors found in the logs".format(error))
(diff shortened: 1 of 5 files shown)


utilities/scan_logs.py
```diff
@@ -68,7 +69,8 @@ def scan_for_errors(log_file_path, error_strings):
             for word in error_strings:
                 # convert the word to lowercase and the line to all lower case
                 # which handles the case where 'warning' will catch 'WARNING' and 'Warning', etc
-                if word.lower() in line.lower():
+                # if word.lower() in line.lower():
+                if re.search(word.lower(), line.lower()):
```
sethrosetter: Can you add some unit tests?

raghusarangapani: It is already there, test_scan_logs.py

raghusarangapani: This change is to enable scanning for strings with regex, like - "Out of memory: Kill process 1234 (sync_gateway)" as "Out of memory: Kill process \d+ \(sync_gateway\)"

sethrosetter: Test for new keywords

testsuites/syncgateway/system/conftest.py
```diff
@@ -180,6 +182,7 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
 
     # Verify all sync_gateways and sg_accels are reachable
     c = cluster.Cluster(cluster_config)
+    remote_executor = RemoteExecutor(c.sync_gateways[0].ip)
```
sethrosetter: I don't think we can rely on this being a single node. Also. RemoteExecutor will not work with Docker. Ansible will be able to.

testsuites/syncgateway/system/conftest.py
```diff
@@ -183,8 +185,45 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
...
     assert len(errors) == 0
+
+    # Scan logs
+    # SG logs for panic, data race
+    # System logs for OOM
+    sys_log_file = "/var/log/messages"
+
+    playbook_vars = {
+        "error_string": "Out of memory: Kill process",
+        "secondary_error_string": "sync_gateway",
```
sethrosetter: Will this catch sg_accel OOM?

PR #1410, opened 2017-09-08
Increase server timeout

raghusarangapani:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Increase the timeout from 30 to 120

### DIFF
--- a/libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
+++ b/libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
@@ -69,7 +69,7 @@
       register: cb_major_version
 
     - name: COUCHBASE SERVER | Wait for node to be listening on port 8091
-      wait_for: port=8091 delay=5 timeout=30
+      wait_for: port=8091 delay=5 timeout=120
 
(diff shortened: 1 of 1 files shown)


libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
```diff
@@ -69,7 +69,7 @@
       register: cb_major_version
 
     - name: COUCHBASE SERVER | Wait for node to be listening on port 8091
-      wait_for: port=8091 delay=5 timeout=30
+      wait_for: port=8091 delay=5 timeout=120
```
sridevi-15: isn't a bug if server does not come up in 30 seconds? 2 minutes will be more , right?

raghusarangapani: It is intermittent, it could be overloaded network/hypervisor. It is possible that it started up before 30s but we timed out trying for the open port due to load. We've had similar issues with SG. We went from 120 to 30 for SG and saw lot of timeout issues. Bumping it back to 120 made all timeouts disappear.

PR #1437, opened 2017-10-18
Feature/mobile opt in

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Added missing method in mobileRest Client and test passes now.


keywords/MobileRestClient.py
```diff
@@ -2051,6 +2051,13 @@ def get_view(self, url, db, design_doc_name, view_name, auth=None):
 
         return resp.json()
 
+    def view_query_through_channels(self, url, db):
```
raghusarangapani: Can you include a doc string? Otherwise looks ok

PR #1463, opened 2017-11-17
Initial install test to validate correct version of SG/A has been installed and is running

ajres:

### DIFF
--- a/testsuites/syncgateway/install/conftest.py
+++ b/testsuites/syncgateway/install/conftest.py
@@ -1,0 +1,382 @@
+""" Setup for Sync Gateway functional tests """
+
+import pytest
+import os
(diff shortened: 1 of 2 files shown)


raghusarangapani:
Hi Andy, in the test and the conftest, all parsing/references to liteserv-host, liteserv-port, liteserv-version, liteserv-platform, liteserv-storage-engine, server-upgraded-version, sync-gateway-upgraded-version and cbs-upgrade-toybuild can be removed. They are not used in this test.

sridevi-15:
To me, this PR is not required. Install and verification of syncgateway version is being testing in our regression . We have upgrade test already under sync gateway which does same. I would say we have to remove this PR

PR #1484, opened 2018-01-29
Add quick start steps

tleyden:
First step in simplifying instructions.


tleyden:
Nevermind, I'm going through the instructions some more, and these steps might be misleading

PR #1512, opened 2018-04-30
new APIs for database encryption and resetCheckPoint

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- New api for database encryption(uncommented) and resetCheckPoint

### DIFF
--- a/CBLClient/Apps/CBLTestServer-iOS/CBLTestServer-iOS/Server/DatabaseConfigurationRequestHandler.swift
+++ b/CBLClient/Apps/CBLTestServer-iOS/CBLTestServer-iOS/Server/DatabaseConfigurationRequestHandler.swift
@@ -23,41 +23,41 @@ public class DatabaseConfigurationRequestHandler {
             
         case "databaseConfiguration_configure":
             let directory: String? = args.get(name:"directory")!
-            // let password: String? = args.get(name:"password")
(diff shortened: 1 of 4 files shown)


CBLClient/Apps/CBLTestServer-iOS/CBLTestServer-iOS/Server/DatabaseConfigurationRequestHandler.swift
```diff
@@ -23,41 +23,41 @@ public class DatabaseConfigurationRequestHandler {
...
             config.directory = directory
             return config
-            
-        /*case "databaseConfiguration_setEncryptionKey":
+
+        case "databaseConfiguration_setEncryptionKey":
             let config: DatabaseConfiguration = args.get(name: "config")!
             let password: String? = args.get(name:"password")!
-            //let encryptionKey: EncryptionKey? = password != nil ? EncryptionKey.password(password!) : nil
+            let encryptionKey: EncryptionKey? = EncryptionKey.password(password!)
```
raghusarangapani: We might want to check for nil password here, else app might crash.

sridevi-15: I remember now that we are already handling nil value on server.swift. If it is nil, it replaces the value with space and never gets nil to the API methods. So we do not require this

PR #1517, opened 2018-05-17
Fix the png deletions under resources/data

raghusarangapani:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Fix the png deletions under resources/data
-

### DIFF
--- a/keywords/attachment.py
+++ b/keywords/attachment.py
@@ -1,7 +1,6 @@
 import base64
 import random
 import uuid
-import os
 
 from PIL import Image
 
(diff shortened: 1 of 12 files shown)


sridevi-15:
changes looks good, just that we might need to add it to all conftest files

raghusarangapani:
Added the call to clear_resources_pngs() from other conftest.py too. Also increased the changestracker timeout for 1 xattrs test, it fixes the test failing randomly.

PR #1541, opened 2018-09-25
Fix/multiple sg tests fix

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- fixed configuration of sync-gateway for topology specific tests
- fixed wrong blip exception handling on ios
- fixed multiple sgs for xamarin -ios

### DIFF
--- a/keywords/TestServeriOS.py
+++ b/keywords/TestServeriOS.py
@@ -53,6 +53,7 @@ class TestServeriOS(TestServerBase):
         if self.platform == "xamarin-ios":
             self.app_dir = "TestServer.iOS/Payload"
             self.package_name = "TestServer.iOS.ipa"
+            self.bundle_id = "com.couchbase.TestServer.iOS"
 
     def download(self, version_build=None):
         """
--- a/keywords/TestServeriOS.py
+++ b/keywords/TestServeriOS.py
@@ -108,10 +108,7 @@ class TestServeriOS(TestServerBase):
         elif self.platform == "xamarin-ios":
             self.app_name = "{}.app".format(self.app)
 
-        if self.platform == "ios":
-            self.app_path = "{}/{}/{}".format(BINARY_DIR, self.app_dir, self.app_name)
-        elif self.platform == "xamarin-ios":
-            self.app_path = "{}/{}/{}".format(BINARY_DIR, self.app_dir, self.app_name)
+        self.app_path = "{}/{}/{}".format(BINARY_DIR, self.app_dir, self.app_name)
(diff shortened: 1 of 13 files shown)


keywords/TestServeriOS.py
```diff
@@ -94,12 +100,18 @@ def install_device(self):
...
+        if self.platform == "ios":
+            if self.debug_mode:
+                self.app_name = "{}-{}-Device-debug.app".format(self.app, self.version_build)
+            else:
+                self.app_name = "{}-{}-Device.app".format(self.app, self.version_build)
+        elif self.platform == "xamarin-ios":
+            self.app_name = "{}.app".format(self.app)
 
-        self.app_path = "{}/{}/{}".format(BINARY_DIR, self.app_dir, self.app_name)
+        if self.platform == "ios":
```
hrajput89: if and elif statement body are same

resources/sync_gateway_configs/listener_tests/multiple_sync_gateways_cc.json
```diff
@@ -21,7 +21,7 @@
             {{ autoimport }}
             {{ xattrs }}
             {{ no_conflicts }}
-            {{ revs_limit }}
+            "revs_limit": 25,
```
hrajput89: why are we hardcoding values here?

resources/sync_gateway_configs/listener_tests/multiple_sync_gateways_cc.json
```diff
@@ -9,7 +9,7 @@
             {{ autoimport }}
             {{ xattrs }}
             {{ no_conflicts }}
-            {{ revs_limit }}
+            "revs_limit": 30,
```
hrajput89: why are we hardcoding values here?

resources/sync_gateway_configs/listener_tests/multiple_sync_gateways_cc.json
sridevi-15: this files are required only for specific tests which needs specific revs limit. As it is multiple sgs, I have a problem of overriding cluster config with revision limit for temporary file as cluster configs temporary files gets override each other and both sync-gateways are having same revision limit at the end which I see issue and change it hard coded with different revision limit

resources/sync_gateway_configs/listener_tests/multiple_sync_gateways_cc.json
sridevi-15: same as above

PR #1551, opened 2018-10-29
Adding skip fix for di mode

hrajput89:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Fix skipping test for di mode

### DIFF
--- a/utilities/cluster_config_utils.py
+++ b/utilities/cluster_config_utils.py
@@ -133,10 +133,7 @@ def is_cbs_ssl_enabled(cluster_config):
 def is_x509_auth(cluster_config):
     ''' Load cluster config to see if auth should be done using x509 certs '''
     cluster = load_cluster_config_json(cluster_config)
-    val = cluster["environment"]["x509_certs"]
-    if val == "True" or val == "true":
-        return True
-    return False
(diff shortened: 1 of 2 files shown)


utilities/cluster_config_utils.py
```diff
@@ -133,7 +133,10 @@ def is_cbs_ssl_enabled(cluster_config):
 def is_x509_auth(cluster_config):
     ''' Load cluster config to see if auth should be done using x509 certs '''
     cluster = load_cluster_config_json(cluster_config)
-    return cluster["environment"]["x509_certs"]
+    val = cluster["environment"]["x509_certs"]
+    if val == "True" or val == "true":
```
sridevi-15: you do not need to add condition here. On conftest file, when it ready pytest parameters, it should assign false as default. Like enabling xattrs. --xattrs will enable true and if --xattrs is not added as parameter, it should assign false by default.

sridevi-15:
I ran this job with the fix you made with you branch and still failing
http://uberjenkins.sc.couchbase.com:8080/job/cen7-sync-gateway-functional-tests-base-di/1583/

PR #1573, opened 2019-02-21
Fix/sg upgrade testing

euniceatcouchbase:
#### Fixes #.

- [X] Ran `flake8`
- [X] Ran `run_repo_tests.sh`

- added missing parameters in sg config file during upgrade
- set use_views value dynamically based on cb server version
- added an utility method to pull cb server version
- added condition check in ansible script in order to conform various cb server version cli command

### DIFF
--- a/keywords/SyncGateway.py
+++ b/keywords/SyncGateway.py
@@ -557,7 +557,7 @@ class SyncGateway(object):
             "no_conflicts": "",
             "revs_limit": "",
             "sg_use_views": "",
-            "num_index_replicas": "",           
+            "num_index_replicas": "",
             "couchbase_server_primary_node": couchbase_server_primary_node
         }
 
--- a/keywords/SyncGateway.py
+++ b/keywords/SyncGateway.py
@@ -646,6 +646,8 @@ class SyncGateway(object):
         ansible_runner = AnsibleRunner(cluster_config)
         server_port = 8091
         server_scheme = "http"
+        sg_cert_path = os.path.abspath(SYNC_GATEWAY_CERT)
+        bucket_names = get_buckets_from_sync_gateway_config(sg_conf)
 
         if is_cbs_ssl_enabled(cluster_config):
             server_port = 18091
@@ -653,6 +655,13 @@ class SyncGateway(object):
 
         # Shared vars
         playbook_vars = {
+            "username": "",
+            "password": "",
+            "certpath": "",
+            "keypath": "",
+            "cacertpath": "",
+            "x509_auth": False,
+            "sg_cert_path": sg_cert_path,
             "sync_gateway_config_filepath": sg_conf,
             "server_port": server_port,
             "server_scheme": server_scheme,
@@ -663,8 +672,12 @@ class SyncGateway(object):
             "sg_use_views": "",
             "revs_limit": "",
             "xattrs": "",
+            "no_conflicts": "",
             "delta_sync": ""
         }
+
+        playbook_vars["username"] = '"username": "{}",'.format(bucket_names[0])
+        playbook_vars["password"] = '"password": "password",'
 
         if get_sg_version(cluster_config) >= "2.1.0":
             logging_config = '"logging": {"debug": {"enabled": true}'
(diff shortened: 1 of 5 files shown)


keywords/SyncGateway.py
```diff
@@ -561,11 +578,16 @@ def upgrade_sync_gateways(self, cluster_config, sg_conf, sync_gateway_version, u
                 log_info("Keyerror in getting logging{}".format(ex.message))
                 playbook_vars["logging"] = '{} {},'.format(logging_config, "}")
 
-            if get_sg_use_views(cluster_config):
-                playbook_vars["sg_use_views"] = '"use_views": true,'
+            server_version = get_cbs_version(cluster_config)
+            cbs_version, cbs_build = version_and_build(server_version)
+            if cbs_version >= "5.5.0":
+                if get_sg_use_views(cluster_config):
+                    playbook_vars["sg_use_views"] = '"use_views": true,'
```
sridevi-15: It is same logic even for else. you can combine the logic

libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
```diff
@@ -128,39 +127,73 @@
       shell: "{{ couchbase_server_home_path }}/bin/couchbase-cli node-init -c {{ couchbase_server_node }}:{{ couchbase_server_admin_port }} --user={{ couchbase_server_admin }} --password={{ couchbase_server_password }} --node-init-hostname={{ couchbase_server_node }}"
       when: "'3.0' in couchbase_server_package_name or '3.1' in couchbase_server_package_name"
 
-    - name: COUCHBASE SERVER | Initialize primary node | configure IPv4
+    - name: COUCHBASE SERVER | Initialize primary node (4.X and 5.X) | configure IPv4
+      shell: "{{ couchbase_server_home_path }}/bin/couchbase-cli node-init -c {{ couchbase_server_node }}:{{ couchbase_server_admin_port }} --user={{ couchbase_server_admin }} --password={{ couchbase_server_password }} --node-init-hostname={{ couchbase_server_node }}"
+      when: "{{ cb_major_version['stdout'] | search('[4-5]')}} and not ipv6_enabled"
```
sridevi-15: can you verify that major version is 15 or 14, it still skips and this condition applies only for 4. or 5. , but not for 15. or 14.

keywords/SyncGateway.py
```diff
@@ -561,11 +578,16 @@ def upgrade_sync_gateways(self, cluster_config, sg_conf, sync_gateway_version, u
...
+                else:
+                    num_replicas = get_sg_replicas(cluster_config)
+                    playbook_vars["num_index_replicas"] = '"num_index_replicas": {},'.format(num_replicas)
             else:
-                num_replicas = get_sg_replicas(cluster_config)
-                playbook_vars["num_index_replicas"] = '"num_index_replicas": {},'.format(num_replicas)
+                playbook_vars["sg_use_views"] = '"use_views": true,'
```
hrajput89: I'm not sure if SG versions <5.5 recognises "use_views" as property. Can you check with Ben or Sri, regarding same? As far as I remember, use_views was added as to 5.5 for backward compatibility.

euniceatcouchbase: comments from Adam: 
SG 2.1 uses GSI by default, which requires CBS 5.5 or later.  The use_views flag was included in SG 2.1 to support running w/ older versions of CBS (by setting use_views=true)
full details here https://docs.couchbase.com/sync-gateway/2.1/compatibility-matrix.html

keywords/SyncGateway.py
euniceatcouchbase: checked in a new copy, combined the logic

libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
euniceatcouchbase: tuned regex pattern to check single digit number if version is 4.x or 5.x and check single digit + not between 1 to 5 if version is 6.x or above

keywords/SyncGateway.py
euniceatcouchbase: done

libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
euniceatcouchbase: done

PR #1583, opened 2019-04-03
Fix/cbl2.5 system test

hrajput89:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Made changes to System test for optional rebalancing and file logging
- Adding file logging options to all conftests


testsuites/CBLTester/CBL_Functional_tests/conftest.py
```diff
@@ -490,6 +503,11 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
     db = None
     cbl_db = None
     if create_db_per_test:
+        if enable_file_logging and liteserv_version >= "2.5.0":
+            cbllog = FileLogging(base_url)
+            cbllog.configure(log_level="verbose", max_rotate_count=2,
```
sridevi-15: can we have something like enable_file_logging=verbose or debug or something else? Whatever is set, it sets log_level? I know you need another verification whether valid log_levels are added or not. We don't really care about other log levels. If it is too hard, you can leave as it is

testsuites/CBLTester/topology_specific_tests/multiple_sync_gateways/conftest.py
```diff
@@ -243,8 +246,6 @@ def params_from_base_suite_setup(request):
             logging_helper.fetch_and_analyze_logs(cluster_config=cluster_config, test_name=request.node.name)
             raise
 
-    cluster.reset(sg_config)
```
sridevi-15: Why is this removed?

testsuites/CBLTester/CBL_Functional_tests/conftest.py
hrajput89: basically verbose is enabled only for debug builds. For release build verbose and debug are same. Hence I provided verbose as log level

testsuites/CBLTester/topology_specific_tests/multiple_sync_gateways/conftest.py
hrajput89: It's not needed as we reset cluster in test itself. Also cluster reset is just after provisioning; doesn't make any sense.

PR #1584, opened 2019-04-03
fix converting blob data type issue in database_getDocuments for all …

euniceatcouchbase:
…platforms

#### Fixes #.

- [X] Ran `flake8`
- [X] Ran `run_repo_tests.sh`

- added code to deal blob type conversion in TestServer app for all platforms. Blob type is used when attachments involved. 
- fixed console log enabling code, the original code was deprecated.


sridevi-15:
test with nonblob scenario too .

PR #1585, opened 2019-04-23
Feature/cbl2.5 20mb doc

hrajput89:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Added test case for 20mb+ push replication

### DIFF
--- a/CBLClient/Apps/CBLTestServer-Android/app/build.gradle
+++ b/CBLClient/Apps/CBLTestServer-Android/app/build.gradle
@@ -28,7 +28,7 @@ android {
     }
 }
 
-final String DEF_VERSION = "2.5.0-220"
+final String DEF_VERSION = "2.5.0-208"
 def COUCHBASE_LITE_VERSION = System.getProperty('version', DEF_VERSION)
 
 dependencies {
--- a/CBLClient/Apps/CBLTestServer-Android/app/build.gradle
+++ b/CBLClient/Apps/CBLTestServer-Android/app/build.gradle
@@ -28,7 +28,7 @@ android {
     }
 }
 
-final String DEF_VERSION = "2.5.0-208"
+final String DEF_VERSION = "2.5.0-220"
 def COUCHBASE_LITE_VERSION = System.getProperty('version', DEF_VERSION)
 
 dependencies {
--- a/CBLClient/Database.py
+++ b/CBLClient/Database.py
@@ -239,7 +239,7 @@ class Database(object):
                 doc_body["channels"] = channels
 
             if attachments_generator:
-                if attachment_file_list!=None:
+                if attachment_file_list is not None:
                     attachments = attachments_generator(attachment_file_list)
                 else:
                     types.verify_is_callable(attachments_generator)
--- a/CBLClient/Database.py
+++ b/CBLClient/Database.py
@@ -209,7 +209,7 @@ class Database(object):
             args.setString("concurrencyControlType", concurrencyControlType)
         return self._client.invokeMethod("database_deleteWithConcurrency", args)
 
-    def create_bulk_docs(self, number, id_prefix, db, channels=None, generator=None, attachments_generator=None, id_start_num=0):
+    def create_bulk_docs(self, number, id_prefix, db, channels=None, generator=None, attachments_generator=None, id_start_num=0, attachment_file_list=None):
         """
         if id_prefix == None, generate a uuid for each doc
 
@@ -239,8 +239,11 @@ class Database(object):
                 doc_body["channels"] = channels
 
             if attachments_generator:
-                types.verify_is_callable(attachments_generator)
-                attachments = attachments_generator()
+                if attachment_file_list!=None:
+                    attachments = attachments_generator(attachment_file_list)
+                else:
+                    types.verify_is_callable(attachments_generator)
+                    attachments = attachments_generator()
                 doc_body["_attachments"] = {att.name: {"data": att.data} for att in attachments}
 
             if id_prefix is None:
--- a/CBLClient/Database.py
+++ b/CBLClient/Database.py
@@ -230,6 +230,8 @@ class Database(object):
                 doc_body = doc_generators.simple_user()
             elif generator == "complex_doc":
                 doc_body = doc_generators.complex_doc()
+            elif generator == "20mb":
+                doc_body = doc_generators.doc_20mb()
             else:
                 doc_body = doc_generators.simple()
 
--- a/libraries/data/doc_generators.py
+++ b/libraries/data/doc_generators.py
@@ -3,6 +3,7 @@ import string
 import random
 import uuid
 import datetime
+import json
 
 
 def random_bool():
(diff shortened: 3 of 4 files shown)


testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
```diff
@@ -484,3 +485,75 @@ def test_replication_delete_event(params_from_base_test_setup, num_of_docs):
...
+
+    replicator = Replication(base_url)
+    repl_config = replicator.configure(source_db=cbl_db,
+                                       target_url=sg_blip_url,
+                                       continuous=False,
+                                       headers=session_header)
+    repl = replicator.create(repl_config)
+    repl_change_listener = replicator.addReplicatorEventChangeListener(repl)
+
+    # 3. Starting Replication and waiting for it finish
```
euniceatcouchbase: should it be #2?

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
```diff
@@ -484,3 +485,75 @@ def test_replication_delete_event(params_from_base_test_setup, num_of_docs):
...
+    replicator.wait_until_replicator_idle(repl, max_times=5000)
+
+    # 4. Getting changes from the replication event listener
+    doc_repl_event_changes = replicator.getReplicatorEventChanges(repl_change_listener)
+    replicator.removeReplicatorEventListener(repl, repl_change_listener)
+    replicator.stop(repl)
+
+    # Processing received events
+    replicated_event_changes = get_event_changes(doc_repl_event_changes)
+    print replicated_event_changes
```
euniceatcouchbase: it's better to use log_info instead of print here, print can be used on local debugging.

libraries/data/doc_generators.py
```diff
@@ -26,6 +27,12 @@ def random_string(length):
     return "".join(random.choice(string.ascii_letters) for _ in xrange(length))
 
 
+def doc_20mb():
+    with open("resources/data/20mb_doc.json") as fh:
```
euniceatcouchbase: check file exists before open?

CBLClient/Database.py
```diff
@@ -230,15 +230,20 @@ def create_bulk_docs(self, number, id_prefix, db, channels=None, generator=None,
...
                 doc_body = doc_generators.simple()
 
             if channels is not None:
                 doc_body["channels"] = channels
 
             if attachments_generator:
-                types.verify_is_callable(attachments_generator)
-                attachments = attachments_generator()
+                if attachment_file_list is not None:
+                    attachments = attachments_generator(attachment_file_list)
```
euniceatcouchbase: what type is attachments_generator? I kind of have problem to imagine what it would be during runtime. I don't see any example in existing code for passing parameters to attachments_generator, all I got were attachments_generator()

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
hrajput89: changed the comment as per no.s

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
hrajput89: removed print, as not needed.

libraries/data/doc_generators.py
hrajput89: actually we want tests to fail if doc doesn't exist. So there is no point on doing error checking as we don't want to handle the exception scenario.

CBLClient/Database.py
hrajput89: attachment generator are functions defined to generate different type of attachment defined in attachment.py. One of function (load_from_data_dir) is to load attachment from a given path for which we didn't provide a provision for in create_bulk docs, so I've added additional argument, attachment_file_list.

CBLClient/Apps/CBLTestServer-Android/app/build.gradle
```diff
@@ -28,7 +28,7 @@ android {
     }
 }
 
-final String DEF_VERSION = "2.5.0-208"
```
sridevi-15: no need to checkin this file. you can add it to the gitignore to avoid checking in

CBLClient/Database.py
sridevi-15: you are attaching one file, does it need to be sent as a list? Can you make it as one file ?

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
```diff
@@ -484,3 +485,75 @@ def test_replication_delete_event(params_from_base_test_setup, num_of_docs):
     doc_ids = db.getDocIds(cbl_db)
     for sg_doc in docs_to_modify:
         assert sg_doc["id"] not in doc_ids, "channel access removal didn't purge the docs from cbl db"
+
+
+@pytest.mark.listener
+@pytest.mark.replication
+@pytest.mark.parametrize("attachment_generator, attachment_file_list", [
+    [None, None],
+    [load_from_data_dir, ["Sample_Image_30mb.jpg"]],
```
sridevi-15: related to same comment mentioned above. Can we have send as one file rather than having a list? Is there any advantage?

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
```diff
@@ -484,3 +485,75 @@ def test_replication_delete_event(params_from_base_test_setup, num_of_docs):
...
+    [generate_2_png_100_100, None],
+])
+def test_push_replication_for_20mb_doc(params_from_base_test_setup, attachment_generator, attachment_file_list):
+    """
+    @summary:
+    1. Create docs in CBL and replicate to SG using push one-shot replication
+    2. Add update to SG and CBL to create conflict
```
sridevi-15: I do not see this step implemented. If this step not required, can you remove it from summary?

CBLClient/Apps/CBLTestServer-Android/app/build.gradle
hrajput89: removed this change.

CBLClient/Database.py
hrajput89: this is function which can be used in different scenario. It is not designed just for one case, so we need to plan it on scalable base.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
hrajput89: the advantage is if in future we want to change and check with multiple files, we don't need to make any change. Program should be written from scalability perspective not just current scenario.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
hrajput89: removed.

PR #1586, opened 2019-04-23
Fix/windows ansible

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- fixed windows tests

### DIFF
--- a/libraries/provision/ansible/playbooks/sgcollect-info.yml
+++ b/libraries/provision/ansible/playbooks/sgcollect-info.yml
@@ -14,9 +14,6 @@
   - include: tasks/sg-sgcollect-info-windows.yml
     when: ansible_os_family == "Windows"
 
-  - include: tasks/sg-pullRedacted-zip-windows.yml
-    when: ansible_os_family == "Windows"
-
 
 - hosts: sg_accels
   become: yes
@@ -32,7 +29,4 @@
     when: ansible_distribution == "CentOS" or ansible_distribution == "RedHat" or ansible_distribution == "Amazon" or ansible_distribution == "Ubuntu" or ansible_distribution == "Debian"
 
   - include: tasks/sa-sgcollect-info-windows.yml
-    when: ansible_os_family == "Windows"
-
-  - include: tasks/sa-pullRedacted-zip-windows.yml
     when: ansible_os_family == "Windows"
--- a/libraries/utilities/generate_clusters_from_pool.py
+++ b/libraries/utilities/generate_clusters_from_pool.py
@@ -404,6 +404,17 @@ def write_config(config, pool_file, use_docker, sg_windows, sg_accel_windows,
             }
         }
 
+        if sg_windows:
+            sg_dict = {
+                "sync_gateways:vars": {
+                    "ansible_user": "Administrator",
+                    "ansible_password": "Membase123",
+                    "ansible_port": 5986,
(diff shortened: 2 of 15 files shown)


libraries/provision/ansible/playbooks/sgcollect-info.yml
```diff
@@ -14,9 +14,6 @@
   - include: tasks/sg-sgcollect-info-windows.yml
     when: ansible_os_family == "Windows"
 
-  - include: tasks/sg-pullRedacted-zip-windows.yml
```
hrajput89: why we removed this task?

libraries/provision/ansible/playbooks/sgcollect-info.yml
```diff
@@ -33,6 +30,3 @@
 
   - include: tasks/sa-sgcollect-info-windows.yml
     when: ansible_os_family == "Windows"
-
-  - include: tasks/sa-pullRedacted-zip-windows.yml
```
hrajput89: same as above

libraries/utilities/generate_clusters_from_pool.py
```diff
@@ -404,6 +404,17 @@ def write_config(config, pool_file, use_docker, sg_windows, sg_accel_windows,
             }
         }
 
+        if sg_windows:
+            sg_dict = {
+                "sync_gateways:vars": {
+                    "ansible_user": "Administrator",
```
hrajput89: We should take this from user. It would be difficult to debug in case username/password is of different combination

libraries/provision/ansible/playbooks/sgcollect-info.yml
sridevi-15: pullling redacted zip file is part of  sg-sgcollect-info-windows.yml, so it is removed.

sridevi-15: pullling redacted zip file is part of  sg-sgcollect-info-windows.yml, so it is removed.

libraries/provision/ansible/playbooks/sgcollect-info.yml
sridevi-15: pullling redacted zip file is part of  sg-sgcollect-info-windows.yml, so it is removed.

libraries/utilities/generate_clusters_from_pool.py
```diff
@@ -407,8 +407,8 @@ def write_config(config, pool_file, use_docker, sg_windows, sg_accel_windows,
         if sg_windows:
             sg_dict = {
                 "sync_gateways:vars": {
-                    "ansible_user": "Administrator",
-                    "ansible_password": "Membase123",
+                    "ansible_user": "FakeUser",
```
hrajput89: I'm confused, how is it different from the previous value? This is still a hardcoded value and will never be overwritten for any other combination of Username and password. If the values are not used anywhere, then we shouldn't even write them in config in first place.

sridevi-15: Hemant, can you check previous code which is above this ? generate clusters per pool is used to generate clusters. we do not pass parameters of username and password when we run this script. we replace with fakeuser/fakepassword with actual username/password using 'sed' command on jenkins config and run this.

PR #1589, opened 2019-05-14
Fix/cbs http fix

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- There is a new change on couchbase server 6. Added http while doing cli call
-


libraries/provision/ansible/playbooks/install-couchbase-server-package.yml
```diff
@@ -17,6 +17,7 @@
     couchbase_server_index_ram: 512
     couchbase_server_cluster_ram: "{{((ansible_memtotal_mb|int)*0.8)|int - 512 }}"
     couchbase_server_admin_port: 8091
+    protocol: http
```
hrajput89: We should not hardcode variable in Ansible script. It should be passed via config or playbook (in our case through python playbook call). This becomes really difficulty to manage in long term.

PR #1592, opened 2019-05-24
Fix/log rotation tests2

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- fixed log rotation tests to work with Centos and windows platforms

### DIFF
--- a/testsuites/syncgateway/functional/tests/test_log_rotation_new.py
+++ b/testsuites/syncgateway/functional/tests/test_log_rotation_new.py
@@ -439,7 +439,10 @@ def test_log_200mb(params_from_base_test_setup, sg_conf_name):
     for log in SG_LOGS:
         status, stdout, stderr = remote_executor.execute("ls /tmp/sg_logs/ | grep {} | wc -l".format(log))
         # A backup file should be created with 200MB
-        assert stdout[0].rstrip() == SG_LOGS_FILES_NUM[log]
+        if (log == "sg_debug" or log == "sg_info") and sg_platform != "windows":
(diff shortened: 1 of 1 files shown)


testsuites/syncgateway/functional/tests/test_log_rotation_new.py
```diff
@@ -431,7 +439,10 @@ def test_log_200mb(params_from_base_test_setup, sg_conf_name):
     for log in SG_LOGS:
         status, stdout, stderr = remote_executor.execute("ls /tmp/sg_logs/ | grep {} | wc -l".format(log))
         # A backup file should be created with 200MB
-        assert stdout[0].rstrip() == SG_LOGS_FILES_NUM[log]
+        if (log == "sg_debug" or log == "sg_info") and sg_platform != "windows":
+            assert int(stdout[0].rstrip()) == int(SG_LOGS_FILES_NUM[log]) + 1
+        else:
+            assert stdout[0].rstrip() == SG_LOGS_FILES_NUM[log]
```
euniceatcouchbase: what's the difference by casting to int or not? does this line also need "assert int(stdout[0].rstrip()) == int(SG_LOGS_FILES_NUM[log])"?

sridevi-15: stdout[0] and SG_LOGS_FILES_NUM[log]  are  strings(number in form of strings). If the number of log files are same, then you do not need to convert to int. As few files have 1 more additional file, I have to convert to int.

PR #1593, opened 2019-05-29
Fix/remove slow server call warning threshold config

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Removed invalid configs of 2.6
- It should effect previous versions too. Ran job on 2.1 and worked fine


sridevi-15:
> Looks good. Also, can we remove di files, as they are no longer supported?

AS di feature just recently removed, let's have it for some time. If we get back , we might need to spend time to modify again. 
Also, let's have di clean up in a separate PR as all tests needs to clean up too which might take some time and need to test all of them once modification is done. We can put it as a task in our task list

PR #1596, opened 2019-06-04
Fix/push replication with attachment

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Added support for create blob and push 
- Added Api, blob methods and images to the app


testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
```diff
@@ -128,6 +145,9 @@ def test_delta_sync_replication(params_from_base_test_setup, num_of_docs, replic
         assert delta_size < doc_writes_bytes, "did not replicate just delta"
 
     sg_docs = sg_client.get_all_docs(url=sg_admin_url, db=sg_db, include_docs=True)["rows"]
+    doc_ids = db.getDocIds(cbl_db)
+    cbl_db_docs = db.getDocuments(cbl_db, doc_ids)
+    print("all cbl docs are ----))))))-----", cbl_db_docs)
```
euniceatcouchbase: is this print only for debug only during coding?

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
```diff
@@ -94,9 +99,21 @@ def test_delta_sync_replication(params_from_base_test_setup, num_of_docs, replic
...
-                doc_body["new_field_2"] = random_string(length=80)
-                db.updateDocument(database=cbl_db, data=doc_body, doc_id=doc_id, attachments_name=file_attachment)
+                if file_attachment:
+                    mutable_dictionary = dictionary.toMutableDictionary(doc_body)
+                    dictionary.setString(mutable_dictionary, "new_field_1", random_string(length=30))
+                    dictionary.setString(mutable_dictionary, "new_field_2", random_string(length=80))
+
+                    if liteserv_platform == "android":
+                        blob_value = blob.create("image/jpeg", blob.createImageContent("/assets/golden_gate_large.jpg"))
+                    elif liteserv_platform == "xamarin-android":
```
euniceatcouchbase: may use file attachment parameter name to construct this, instead of manually have the file name "golden_gate_large.jpg"

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
```diff
@@ -94,9 +99,21 @@ def test_delta_sync_replication(params_from_base_test_setup, num_of_docs, replic
...
+                        image_content = blob.createImageContent("Assets/golden_gate_large.jpg")
+                        blob_value = blob.create("image/jpeg", stream=image_content)
+                    else:
+                        image_content = blob.createImageContent("Files/golden_gate_large.jpg")
```
euniceatcouchbase: same as above

CBLClient/Database.py
```diff
@@ -121,10 +121,8 @@ def updateDocuments(self, database, documents):
         args.setDictionary("documents", documents)
         return self._client.invokeMethod("database_updateDocuments", args)
 
-    def updateDocument(self, database, data, doc_id, attachments_name=None):
+    def updateDocument(self, database, data, doc_id):
```
hrajput89: what happens to the case in which we want to update attachment?

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
hrajput89: ditto

sridevi-15: There is a reason for not using it. These golden_gate image should store it in TestServer app, so that when it loads the app, it can find. It is kind of hard coded as only one image exists in app. If you parameterized and later if you change the parameter to some other image, this will fail saying image not found .

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
sridevi-15: same as above

CBLClient/Database.py
sridevi-15: we are using blobs now, there is no attachment in CBL. When you update document, data should have image added and image comes along with data. There is no need of attachment parameter as we don't use attachments. I made sure it is not used in any tests.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
sridevi-15: done, removed the print statement. Added for debugging, forgot to remove

CBLClient/Database.py
hrajput89: If we don't have a test which doesn't update attachment, then we should write one. This is miss on our part that we are not testing anywhere, how update behaves if we just update attachment but not content

sridevi-15: already added @hemant, that is the whole reason for adding blob support as test is failing.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
```diff
@@ -94,9 +99,24 @@ def test_delta_sync_replication(params_from_base_test_setup, num_of_docs, replic
         cbl_db_docs = db.getDocuments(cbl_db, doc_ids)
         for doc_id, doc_body in cbl_db_docs.items():
             for i in range(number_of_updates):
-                doc_body["new_field_1"] = random_string(length=30)
-                doc_body["new_field_2"] = random_string(length=80)
-                db.updateDocument(database=cbl_db, data=doc_body, doc_id=doc_id, attachments_name=file_attachment)
+                if file_attachment:
+                    mutable_dictionary = dictionary.toMutableDictionary(doc_body)
+                    dictionary.setString(mutable_dictionary, "new_field_1", random_string(length=30))
```
hrajput89: all set methods return Mutable Dictionary object, however, you are nowhere retaining that object. Does object from toMutableDictionary call retrains all future changes?

This is for all set methods?

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
```diff
@@ -94,9 +99,24 @@ def test_delta_sync_replication(params_from_base_test_setup, num_of_docs, replic
...
+                        image_content = blob.createImageContent("golden_gate_large.jpg")
+                        blob_value = blob.create("image/jpeg", stream=image_content)
+                    elif liteserv_platform == "ios":
+                        image_content = blob.createImageContent("Files/golden_gate_large.jpg")
+                        blob_value = blob.create("image/jpeg", content=image_content)
+                    else:
+                        image_content = blob.createImageContent("Files/golden_gate_large.jpg")
+                        blob_value = blob.create("image/jpeg", stream=image_content)
+                    dictionary.setBlob(mutable_dictionary, "_attachments", blob_value)
+                db.updateDocument(database=cbl_db, data=doc_body, doc_id=doc_id)
```
hrajput89: I see that you created a new mutable_dictionary if there is new attachment update, however, you didn't pass it through. You are still using doc_body, which has same value as before update or I'm missing something?

sridevi-15: Good question. When we setBlob, it effects the doc_body directly . I had same question and I tested manually too to make sure doc_body after updating doc has blob and new fields.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_delta_sync.py
sridevi-15: When I set values, it effects doc_body, I do not require mutable dictionary . For update document, I need doc_body. Also compare docs assertion will make sure new fields and attachments are replicated which proves that new changes are effected.

PR #1600, opened 2019-07-19
CBL-239: Add retrieving log endpoint. The response can be either regular text f…

Sandychuang8:
Add retrieving log endpoint. The response can be either regular text format or binary format. Need to enable the file logger first in order to get the logs, otherwise the response will be an empty string.

### DIFF
--- a/CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
+++ b/CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Net;
 using JetBrains.Annotations;
 using Couchbase.Lite.Logging;
+using System.IO.Compression;
 
 namespace Couchbase.Lite.Testing
 {
@@ -156,7 +157,7 @@ namespace Couchbase.Lite.Testing
 
         }
 
-        public static void GetLogsInByteArray([NotNull] NameValueCollection args,
+        public static void GetLogsInZip([NotNull] NameValueCollection args,
                                 [NotNull] IReadOnlyDictionary<string, object> postBody,
                                 [NotNull] HttpListenerResponse response)
         {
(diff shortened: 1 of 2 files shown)


sridevi-15:
changes are good to me. I want to use same branch to add this API to our python test , verify tests are working without breaking and then we can merge to master

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
```diff
@@ -156,6 +157,42 @@ public class FileLoggingMehtod
 
         }
 
+        public static void GetLogsInZip([NotNull] NameValueCollection args,
+                                [NotNull] IReadOnlyDictionary<string, object> postBody,
+                                [NotNull] HttpListenerResponse response)
+        {
+            var path = Database.Log.File.Config?.Directory;
```
hrajput89: We create separate logs for each test, so we need to pass directory for this purpose. Change this code to accept the path string provided by user.

I've used below change for my testing purpose

var path = postBody["path"].ToString();

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
```diff
@@ -156,6 +157,42 @@ public class FileLoggingMehtod
...
+                    using (var entryStream = entry.Open()) {
+                        file.CopyTo(entryStream);
+                    }
+                }
+            }
+
+            try {
+                response.ContentLength64 = memoryStream.Length;
+                Stream output = response.OutputStream;
+                memoryStream.WriteTo(output);
```
hrajput89: does this means that method is sending binary data on the channel and user has to write it back into a file and save  it as zip at client side?

borrrden: Yes that is what it means

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
borrrden: You mean that by the time you call this method, the current log file directory will no longer be valid?

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
```diff
@@ -156,6 +157,42 @@ public class FileLoggingMehtod
...
+            if (path == null) {
+                response.WriteEmptyBody();
+                return;
+            }
+
+            string[] filePaths = Directory.GetFiles(@path);
+            var memoryStream = new MemoryStream();
+            using (var archive = new ZipArchive(memoryStream, ZipArchiveMode.Create, true)) {
+                foreach (var filePath in filePaths) {
+                    var entry = archive.CreateEntry(Path.GetFileName($"{filePath}.zip"));
```
hrajput89: No need to add .zip extension for log files. Should be removed

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/FileLoggingMehtod.cs
hrajput89: for most of the cases what you say is true, but I remember we have test cases where we have created more than 1 db. It's not giving any issue for now. Will comeback to it later if it gives issue for that case

PR #1604, opened 2019-08-01
Adding delta-sync option to system test conftest

hrajput89:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Added delta-sync option for system-test contest

### DIFF
--- a/testsuites/CBLTester/System_test_multiple_device/conftest.py
+++ b/testsuites/CBLTester/System_test_multiple_device/conftest.py
@@ -134,6 +134,7 @@ def pytest_addoption(parser):
                      action="store_true",
                      help="delta-sync: Enable delta-sync for sync gateway")
 
+
 # This will get called once before the first test that
 # runs with this as input parameters in this file
 # This setup will be called once for all tests in the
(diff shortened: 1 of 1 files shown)


testsuites/CBLTester/System_test_multiple_device/conftest.py
```diff
@@ -358,7 +369,8 @@ def params_from_base_suite_setup(request):
         "create_db_per_test": create_db_per_test,
         "enable_rebalance": enable_rebalance,
         "enable_encryption": enable_encryption,
-        "encryption_password": encryption_password
+        "encryption_password": encryption_password,
+        "delta_sync_enabled": delta_sync_enabled
```
euniceatcouchbase: do you need to pass the delta_sync_enabled parameter to test case level?

hrajput89: Not now, but if someone need the flag to skip a test or to do some other check, they would need it

sridevi-15: that's okay for  now, I would suggest not to add unless it is required for the existing tests. We should add as and when required.

PR #1606, opened 2019-08-05
Fix/cbl upgarde 2.6

hrajput89:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

-  Fixed CBL upgrade tests failure for Query tests after upgrade


testsuites/CBLTester/System_test_multiple_device/conftest.py
```diff
@@ -283,6 +282,7 @@ def params_from_base_suite_setup(request):
             raise
 
     # Create CBL databases on all devices
+    sg_admin_url = cluster_topology["sync_gateways"][0]["admin"]
```
euniceatcouchbase: is this a code change or just moved some lines down? the comment above this line doesn't sound matching the logic this line.

sridevi-15: actually there is no difference. Hemant just moved from one line to another

testsuites/CBLTester/upgrade_tests/test_cbl_upgrade.py
```diff
@@ -1,34 +1,37 @@
...
+from keywords.utils import log_info
+from libraries.testkit.cluster import Cluster
+from testsuites.CBLTester.CBL_Functional_tests.SuiteSetup_FunctionalTests.test_query import test_get_doc_ids, \
+    test_any_operator, test_doc_get, test_get_docs_with_limit_offset, test_multiple_selects, test_query_where_and_or, \
+    test_query_pattern_like, test_query_pattern_regex, test_query_isNullOrMissing, test_query_ordering, \
+    test_query_substring, test_query_collation, test_query_join, test_query_inner_join, test_query_cross_join, \
+    test_query_left_join, test_query_left_outer_join, test_equal_to, test_not_equal_to, test_greater_than, \
+    test_greater_than_or_equal_to, test_less_than, test_less_than_or_equal_to, test_in, test_between, test_is, \
+    test_isnot, test_not, test_single_property_fts, test_multiple_property_fts, test_fts_with_ranking, \
+    test_getDoc_withValueTypeDouble, test_getDoc_withLocale, test_query_arthimetic
```
sridevi-15: looks like when new query test added, we should add it here, right?

testsuites/CBLTester/upgrade_tests/test_cbl_upgrade.py
```diff
@@ -94,54 +97,135 @@ def test_upgrade_cbl(params_from_base_suite_setup):
...
+            query_test_func(*args)
+            log_info("PASSED")
+            tests_result["{}{}".format(query_test_func.__name__, args[1:])] = "PASSED"
+            test_passed += 1
+        except Exception as err:
+            log_info("FAILED:\n {}".format(err))
+            tests_result["{}{}".format(query_test_func.__name__, args[1:])] = "FAILED"
+            test_failed += 1
+
+    log_info("\n\nTests Result: PASSED {}, FAILED {}".format(test_passed, test_failed))
```
sridevi-15: if any query test failed, will it show it in jenkins which tests failed? If not, please make upgrade test failed, even 1 test failed . You can add assert(false) if number of failed tests is not zero

testsuites/CBLTester/System_test_multiple_device/conftest.py
hrajput89: Actually no, the reason I moved it because I was initialising the sg_admin_url, before if condition for sg_ssl flag. This was causing the issue for sg_client.create_user call. So basically, admin url get initialised to "http", but after checking for sg_ssl flag, the config file may or may not change for ssl value and thus in case of True, we still use http instead of https, leading to error.

testsuites/CBLTester/upgrade_tests/test_cbl_upgrade.py
hrajput89: Yes, we have to add the new query test in the list. I couldn't find an easier way, so I had to hard code it

testsuites/CBLTester/upgrade_tests/test_cbl_upgrade.py
hrajput89: Actually, you'll the see the result of each query test failure. I've added the logic to mark test failed if any query tests failed.

PR #1607, opened 2019-08-05
fix the missing parameter, and add a flag to allow use locally build …

euniceatcouchbase:
…testserver

#### Fixes #.

- [X] Ran `flake8`
- [X] Ran `run_repo_tests.sh`

- added the missing attribute liteserv_version 
- introduced a new flag --use-local-testserver, which allows to bypass installing package to testserver, easy for debugging, no longer need to comment out the testserver install and stop code blocks.

### DIFF
--- a/testsuites/CBLTester/CBL_Functional_tests/conftest.py
+++ b/testsuites/CBLTester/CBL_Functional_tests/conftest.py
@@ -359,7 +359,7 @@ def params_from_base_suite_setup(request):
         raise Exception("enable_sample_bucket has to be used with create_db_per_suite")
 
     # Start Test server which needed for suite level set up like query tests
-    if (not use_local_testserver) and create_db_per_suite:
+    if not use_local_testserver and create_db_per_suite:
         log_info("Starting TestServer...")
         test_name_cp = test_name.replace("/", "-")
         if device_enabled:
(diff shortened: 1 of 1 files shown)


testsuites/CBLTester/CBL_Functional_tests/conftest.py
```diff
@@ -530,14 +538,15 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
...
+    use_local_testserver = request.config.getoption("--use-local-testserver")
 
     source_db = None
     test_name_cp = test_name.replace("/", "-")
     log_filename = "{}/logs/{}-{}-{}.txt".format(RESULTS_DIR, type(testserver).__name__,
                                                  test_name_cp,
                                                  datetime.datetime.now())
 
-    if create_db_per_test:
+    if (not use_local_testserver) and create_db_per_test:
```
sridevi-15: no need to use brackets here

testsuites/CBLTester/CBL_Functional_tests/conftest.py
```diff
@@ -352,7 +359,7 @@ def params_from_base_suite_setup(request):
         raise Exception("enable_sample_bucket has to be used with create_db_per_suite")
 
     # Start Test server which needed for suite level set up like query tests
-    if create_db_per_suite:
+    if (not use_local_testserver) and create_db_per_suite:
```
sridevi-15: no need to add brackets here

testsuites/CBLTester/CBL_Functional_tests/conftest.py
euniceatcouchbase: made a change based on the feedback

testsuites/CBLTester/CBL_Functional_tests/conftest.py
euniceatcouchbase: made a change based on the feedback

PR #1609, opened 2019-08-08
Fix/remove slow server call warning threshold config

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- modified and replaced with new configs which is valid for cobalt
-fixed tests with new changes in cobalt

### DIFF
--- a/testsuites/syncgateway/functional/tests/test_db_online_offline.py
+++ b/testsuites/syncgateway/functional/tests/test_db_online_offline.py
@@ -461,7 +461,7 @@ def test_online_to_offline_changes_feed_controlled_close_longpoll(params_from_ba
     with concurrent.futures.ThreadPoolExecutor(max_workers=libraries.testkit.settings.MAX_REQUEST_WORKERS) as executor:
         futures = dict()
         futures[executor.submit(seth.start_longpoll_changes_tracking, termination_doc_id=None)] = "polling"
-        #  For windows, requiest is too slow, so have _offline request to start before creating docs
+        #  For windows, request is too slow, so have _offline request to start before creating docs
         if sg_platform == "windows":
             futures[executor.submit(sg_client.take_db_offline, cluster_conf, "db")] = "db_offline_task"
         time.sleep(5)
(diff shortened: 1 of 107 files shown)


testsuites/syncgateway/functional/tests/test_db_online_offline.py
```diff
@@ -460,9 +461,13 @@ def test_online_to_offline_changes_feed_controlled_close_longpoll(params_from_ba
     with concurrent.futures.ThreadPoolExecutor(max_workers=libraries.testkit.settings.MAX_REQUEST_WORKERS) as executor:
         futures = dict()
         futures[executor.submit(seth.start_longpoll_changes_tracking, termination_doc_id=None)] = "polling"
-        futures[executor.submit(doc_pusher.add_docs, num_docs)] = "docs_push"
+        #  For windows, requiest is too slow, so have _offline request to start before creating docs
+        if sg_platform == "windows":
+            futures[executor.submit(sg_client.take_db_offline, cluster_conf, "db")] = "db_offline_task"
         time.sleep(5)
-        futures[executor.submit(sg_client.take_db_offline, cluster_conf, "db")] = "db_offline_task"
```
sridevi-15: made this change from 463 to 465 to have test run successfully on windows. Request to windows is too slow, I have to make dboffline request first and then add docs, otherwise it won't work . For centos, adddocs should be called first and then dboffline

testsuites/syncgateway/functional/tests/test_db_online_offline.py
```diff
@@ -460,9 +461,13 @@ def test_online_to_offline_changes_feed_controlled_close_longpoll(params_from_ba
...
+        futures[executor.submit(doc_pusher.add_docs, num_docs)] = "docs_push"
+        if sg_platform != "windows":
+            futures[executor.submit(sg_client.take_db_offline, cluster_conf, "db")] = "db_offline_task"
```
euniceatcouchbase: is db_offline_task need to run twice (before and after docs_push? if so, please add comments to explain it does need to be run twice, otherwise, one of them needs to be removed.

PR #1619, opened 2019-09-04
Fix/serverpool

hrajput89:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Methods to request/release nodes for Jenkins job


utilities/mobile_server_pool.py
```diff
@@ -0,0 +1,227 @@
...
+
+            pool_list.append(str(doc_id))
+        else:
+            continue
+        if len(pool_list) == int(num_of_nodes):
+            return pool_list
+
+    # Not able to allocate all the requested nodes, hence release the node back to the pool
+    release_node(pool_list, job_name_requesting_node)
+    raise Exception("Not enough free node available to satisfy the request")
```
sridevi-15: you can replace the node with node/nodes as sometime it can be multiple nodes

utilities/mobile_server_pool.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def reserve_node(doc_id, job_name, counter=0):
+    """
+    Reserve a node for a given job
```
sridevi-15: summary explanation is good

utilities/mobile_server_pool.py
```diff
@@ -0,0 +1,227 @@
...
+    :return: list of nodes reserved
+    """
+    # Getting list of available nodes through n1ql query
+    query_str = "select meta().id from `{}` where os='{}' " \
+                "AND os_version='{}' AND state='available'".format(BUCKET_NAME, nodes_os_type, node_os_version)
+    query = N1QLQuery(query_str)
+    pool_list = []
+    for row in sdk_client.n1ql_query(query):
+        doc_id = row["id"]
+        node = reserve_node(doc_id, job_name_requesting_node)
```
euniceatcouchbase: reserve_node function return a boolean, my first impression of the variable name node is that reserve_node return an object (of a machine as a node). it will be nice to rename node to something more meaningful, but this is just a minor comment.

utilities/mobile_server_pool.py
```diff
@@ -0,0 +1,227 @@
...
+        if node:
+
```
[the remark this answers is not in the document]
hrajput89: done

utilities/mobile_server_pool.py
hrajput89: done

utilities/mobile_server_pool.py
hrajput89: I can return void and raise exception if nodes are not reserve. Let me know if that make sense.

euniceatcouchbase: I am fine with the definition of reserve_node and its boolean return type. However, when I read this line of code (without checking the definition of reserve_node), I got a wrong impression that reserve_node returns a node object, instead of a boolean status. You may change the variable "node" to "is_node_available" or something like "nodes_reserve_status", people will not misunderstand what is doing here.

PR #1622, opened 2019-09-05
Fix/cm-174 predictive queries basic input output

euniceatcouchbase:
#### Fixes cm-174.

- [X] Ran `flake8`
- [X] Ran `run_repo_tests.sh`

[requirements]
predictive query results comparison need tolerant for small diff between long or float numbers. such as randomly generated float number 60082543713568.90625 may become a long number like 60082543713613 showing on predictive query result set. They are considered equal in this scenario. 
[fixes]
- add a flag to all compare functions with default value of isPredictiveResult=False
- implemented a tolerant compare for predictive result
- set the flag to True in predictive test case

### DIFF
--- a/keywords/utils.py
+++ b/keywords/utils.py
@@ -4,6 +4,7 @@ import os
 import random
 import string
 import re
+import pdb
 
 from keywords.exceptions import FeatureSupportedError
 from keywords.constants import DATA_DIR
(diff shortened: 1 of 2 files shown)


keywords/utils.py
```diff
@@ -323,7 +323,10 @@ def compare_generic_types(object1, object2):
     elif isinstance(object1, int) and isinstance(object2, long):
         return object1 == int(object2)
     elif isinstance(object1, float) and isinstance(object2, long):
-        return object1 == float(object2)
+        if isPredictiveResult:
```
hrajput89: I have no idea what are you trying to do when isPredictiveResult flag is true and why? The logic is not intuitive as the function name doesn't suggest anything regarding comparison for Predictive results. Everywhere else, we are comparing for equality, but in predictive result we are comparing for less than, that to be of difference.

I'd suggest that you either add a summary in the function doc string or at least add comment around if comparison for predictive result.

euniceatcouchbase: I have added comments to the function definition and where the function is been called. Hope this makes the code more maintainable

PR #1629, opened 2019-10-03
Fix/sgw fixes during mercury part2

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- fixed keyword error for sg_platform
- fixed the import_enabled test to reduce the running time for 30 mins to 10 mins

### DIFF
--- a/testsuites/CBLTester/CBL_Functional_tests/conftest.py
+++ b/testsuites/CBLTester/CBL_Functional_tests/conftest.py
@@ -320,6 +320,9 @@ def params_from_base_suite_setup(request):
         log_info("Running without delta sync")
         persist_cluster_config_environment_prop(cluster_config, 'delta_sync_enabled', False)
 
+    # As cblite jobs run with on Centos platform, adding by default centos to environment config
+    persist_cluster_config_environment_prop(cluster_config, 'sg_platform', "centos", False)
+
     # Write the number of replicas to cluster config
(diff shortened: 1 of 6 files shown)


testsuites/CBLTester/CBL_Functional_tests/conftest.py
```diff
@@ -320,6 +320,9 @@ def params_from_base_suite_setup(request):
         log_info("Running without delta sync")
         persist_cluster_config_environment_prop(cluster_config, 'delta_sync_enabled', False)
 
+    # As cblite jobs run with on Centos platform, adding by default centos to environment config
+    persist_cluster_config_environment_prop(cluster_config, 'sg_platform', "centos", False)
```
hrajput89: don't we have SG jobs for other platforms like ubuntu and macOS?

sridevi-15: > don't we have SG jobs for other platforms like ubuntu and macOS?

we have SG jobs which has  parameter for sg_platform for sg conftest file. This conftest is for CBL which we are not having sg-platform as a parameter and running default on centos. That is why I added default as centos.  For new changes of mac support needs sg_platform parameter in cluster configs 'reset' method expects to have sync-gateway path for x509 changes to work.  So, that mac support can work with x509 changes.

PR #1640, opened 2019-10-31
Fix/cm-303 security policy support for Xamarin-Android

hrajput89:

- Added security policy support for Xamarin-Android for Android Pie(9.0)

Followed the process from here - https://devblogs.microsoft.com/xamarin/cleartext-http-android-network-security/

### DIFF
--- a/CBLClient/Apps/CBLTestServer-Dotnet/TestServer.Android/Resources/xml/network_security_config.xml
+++ b/CBLClient/Apps/CBLTestServer-Dotnet/TestServer.Android/Resources/xml/network_security_config.xml
@@ -1,0 +1,1 @@
+﻿<?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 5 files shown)


CBLClient/Apps/CBLTestServer-Dotnet/TestServer.Android/Resources/xml/network_security_config.xml
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+  <domain-config cleartextTrafficPermitted="true">
+    <domain includeSubdomains="true">10.0.2.2</domain> <!-- Debug port -->
+    <domain includeSubdomains="true">xamarin.com</domain>
```
euniceatcouchbase: have you got a chance to test our xamarin-android testserver app on an Android P device? I doubt this configuration will work, the ip address or domain names suppose to be the sync gateway ip addresses that testserver needs to reach. since in our scenario, sync gateway servers are randomly from the pool, there is no way to list a range of ip addresses, therefore I specified cleartestTrafficPermitted=true in base-config instead of domain-config.

euniceatcouchbase: https://github.com/couchbaselabs/mobile-testkit/blob/00d5b14d3a1f60b5cf33c188120376af7df79aeb/CBLClient/Apps/CBLTestServer-Android/app/src/main/res/xml/network_security_config.xml

PR #1659, opened 2019-12-17
Prevent running setup.sh

bmeike:
setup.sh does the wrong thing if it is run (`./setup.sh`) instead of being sourced (`. setup.sh`).
1) change permissions so that it cannot be executed
2) add check for execution vs sourceing


bmeike:
I don't know why it doesn't show up in the diffs.  It now shows as -rw-r--r-- on my machine.  I think that is what's committed...

PR #1672, opened 2020-02-21
Fix/dotnet systemtest norelease memory

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- avoid release memory for query results for .Net
- job ran on jenkins for .Net successfully
http://uberjenkins.sc.couchbase.com:8080/view/03%20CBL-2.x/job/CBLITE_SYNCGATEWAY_Syt-DotNet/46/
-

### DIFF
--- a/testsuites/CBLTester/System_test_multiple_device/test_system_testing.py
+++ b/testsuites/CBLTester/System_test_multiple_device/test_system_testing.py
@@ -40,6 +40,7 @@ def test_system(params_from_base_suite_setup):
     num_of_docs_to_add = params_from_base_suite_setup["num_of_docs_to_add"]
     up_time = params_from_base_suite_setup["up_time"]
     repl_status_check_sleep_time = params_from_base_suite_setup["repl_status_check_sleep_time"]
+    platform_list = params_from_base_suite_setup["platform_list"]
     doc_id_for_new_docs = num_of_docs
     query_limit = 1000
     query_offset = 0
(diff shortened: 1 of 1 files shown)


testsuites/CBLTester/System_test_multiple_device/test_system_testing.py
```diff
@@ -301,7 +317,9 @@ def test_system(params_from_base_suite_setup):
             t.join()
             results = query.query_get_docs_limit_offset(cbl_db, limit=query_limit,
                                                         offset=query_offset)
-            _releaseQueryResults(base_url, results)
+            # Query results do not store in memory for dot net, so no need to release memory for dotnet
+            if(platform.lower() != "net-msft" and platform.lower() != "uwp" and platform.lower() != "xamarin-ios" and platform.lower() != "xamarin-android"):
```
mgcouchbase: I think you can add the if condition in the method utils.release(results), that way you can just add the condition only once

sridevi-15: We are not supposed to release the objects if we are not storing in memory.  It might not catch issues if we are storing in memory explicitly and if there is a bug and not storing in memory.

mgcouchbase: if we keep all memory related conditions in the  utils.release method it will be easy for future reference and if we see the same issue in new upcoming platform easy to use.

PR #1677, opened 2020-03-16
Fix/separate ios background app

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- isolated backgroud app tests
- fixed iOS-sim command which is failing with simulator

### DIFF
--- a/keywords/TestServeriOS.py
+++ b/keywords/TestServeriOS.py
@@ -174,8 +174,11 @@ class TestServeriOS(TestServerBase):
 
         log_info("Installing: {}".format(self.app_path))
         # Launch the simulator and install the app
+        # output = subprocess.check_output([
+        #     "ios-sim", "--devicetypeid", self.device, "install", self.app_path, "--exit"
+        # ])
         output = subprocess.check_output([
-            "ios-sim", "--devicetypeid", self.device, "install", self.app_path, "--exit"
+             "ios-sim", "install", self.app_path, "--devicetypeid", self.device, "--exit"
         ])
 
         log_info(output)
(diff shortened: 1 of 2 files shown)


keywords/TestServeriOS.py
```diff
@@ -174,8 +174,11 @@ def install(self):
 
         log_info("Installing: {}".format(self.app_path))
         # Launch the simulator and install the app
+        # output = subprocess.check_output([
+        #     "ios-sim", "--devicetypeid", self.device, "install", self.app_path, "--exit"
+        # ])
         output = subprocess.check_output([
-            "ios-sim", "--devicetypeid", self.device, "install", self.app_path, "--exit"
+             "ios-sim", "install", self.app_path, "--devicetypeid", self.device, "--exit"
```
mgcouchbase: Looks like there is no change here

keywords/TestServeriOS.py
```diff
@@ -274,8 +277,11 @@ def start(self, logfile_name):
         # Without --exit, ios-sim blocks
         # With --exit, --log has no effect
         # subprocess.Popen didn't launch the app
+        # output = subprocess.check_output([
+        #    "ios-sim", "--devicetypeid", self.device, "launch", self.app_path, "--exit"
+        # ])
         output = subprocess.check_output([
-            "ios-sim", "--devicetypeid", self.device, "launch", self.app_path, "--exit"
```
mgcouchbase: Changing the order of params would fix the issue ?

sridevi-15: yes Manasa,

keywords/TestServeriOS.py
sridevi-15: There is a change. seeing issues to get device id when ios-sim installed by brew, so did yum install and that needs this order change, otherwise it won't work

PR #1679, opened 2020-03-30
maven url change

mgcouchbase:
#### Fixes #.

- [ ] Ran `flake8`
- [ ] Ran `run_repo_tests.sh`

- 
-


sridevi-15:
There are 2 other places too on java-desktop and java we server too. Can you please modify on that too? just search for that ip in the testkit and replace with mobile.maven.couchbase.com

PR #1692, opened 2020-04-27
Fix/android functional test job

euniceatcouchbase:
#### Fixes #. fix 2 android functional test job issues

- [X] Ran `flake8`
- [X] Ran `run_repo_tests.sh`

- fixed android resource file path issue
- fixed an issue to deal with negative value of float type

### DIFF
--- a/CBLClient/ValueSerializer.py
+++ b/CBLClient/ValueSerializer.py
@@ -25,7 +25,7 @@ class ValueSerializer(object):
             bool_val = bool(value)
             return "true" if bool_val else "false"
         elif isinstance(value, int):
-            if value < 1000000:
+            if value < 1000000 and value > -1000000:
                 number = int(value)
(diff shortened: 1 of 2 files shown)


CBLClient/ValueSerializer.py
```diff
@@ -25,7 +25,7 @@ def serialize(value):
             bool_val = bool(value)
             return "true" if bool_val else "false"
         elif isinstance(value, int):
-            if value < 1000000:
+            if value < 1000000 and value > -1000000:
```
mgcouchbase, line 28: this change require functional test testing

euniceatcouchbase: I ran android functional test job with this branch. will run an ios functional job and 1 .net job with this branch too to verify.

I don't know why this is not an issue in the past, kind of related to the random function behavior change in python3.

PR #1694, opened 2020-04-29
change download location to be align with dev/build changes

euniceatcouchbase:
#### Fixes #. a quick fix for android testserver app output location update

- [X] Ran `flake8`
- [X] Ran `run_repo_tests.sh`

- now android ce and ee builds are merged into one united directory.

### DIFF
--- a/keywords/TestServerAndroid.py
+++ b/keywords/TestServerAndroid.py
@@ -18,12 +18,11 @@ class TestServerAndroid(TestServerBase):
         self.platform = platform
 
         if self.platform == "android":
+            self.download_source = "couchbase-lite-android"
             if community_enabled:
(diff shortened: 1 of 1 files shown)


keywords/TestServerAndroid.py
```diff
@@ -18,12 +18,11 @@ def __init__(self, version_build, host, port, community_enabled=None, debug_mode
         self.platform = platform
 
         if self.platform == "android":
+            self.download_source = "couchbase-lite-android"
             if community_enabled:
                 apk_name_prefix = "CBLTestServer-Android-{}-community".format(self.version_build)
```
mgcouchbase, line 23: CBL do not support the community version right ?

euniceatcouchbase: so far we don't build/test cc version, but this code supports cc from day 1, I'd prefer to keep it

PR #1697, opened 2020-05-04
Fix/getnodes for oscertify

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- added support to share vms for os certification with server team
- avoided the dependency of unzip package and added support to sync the clocks of server and sync gateway
- added project 100 feature on mobile pool script to check the sshability of the nodes 

Due to frame work changes:  ran differrent set of sqw jobs:
Test failures are comparable with the branch without these changes and they are expected. 

SGW functional tests : http://uberjenkins.sc.couchbase.com:8080/job/cen7-sync-gateway-functional-tests-xattrs-userViews-ServerSsl/79/testReport/
upgrade job :  http://uberjenkins.sc.couchbase.com:8080/job/cen7-sync-gateway-upgrade/30/testReport/
upgrade test server job :  http://uberjenkins.sc.couchbase.com:8080/job/sync-gateway-upgrade-centos-testserver/488/testReport/
Windows sgw tests : http://uberjenkins.sc.couchbase.com:8080/job/win10-sync-gateway-functional-tests-base-cc/512/testReport/
Macos SGW tests : http://uberjenkins.sc.couchbase.com:8080/view/os-certifications/job/Macos-sync-gateway-functional-tests-base/67/testReport/
shared centos 8 VM for SGW : http://uberjenkins.sc.couchbase.com:8080/view/os-certifications/job/cen8-sync-gateway-functional-tests/45/testReport/
shared ubuntu 18 vm for SGW : http://uberjenkins.sc.couchbase.com:8080/view/os-certifications/job/ubuntu18-sync-gateway-functional-tests/10/testReport/

### DIFF
--- a/utilities/cluster_config_utils.py
+++ b/utilities/cluster_config_utils.py
@@ -103,20 +103,20 @@ def generate_x509_certs(cluster_config, bucket_name, sg_platform):
     stdout, stderr = proc.communicate()
     print(stdout, stderr)
 
-    # zipping the certificates
+    # # zipping the certificates
     os.chdir(curr_dir)
-    make_archive("certs", "zip", certs_dir)
+    # make_archive("certs", "zip", certs_dir)
 
-    for node in cluster["sync_gateways"]:
-        if sg_platform.lower() != "macos" and sg_platform.lower() != "windows":
-            cmd = ["scp", "certs.zip", "{}@{}:/tmp".format(username, node["ip"])]
-        else:
-            cmd = ["cp", "certs.zip", "/tmp"]
-        print(" ".join(cmd))
-        proc = Popen(cmd, stdout=PIPE, stderr=PIPE)
-        stdout, stderr = proc.communicate()
-        if stdout or stderr:
-            print(stdout, stderr)
+    # for node in cluster["sync_gateways"]:
+    #     if sg_platform.lower() != "macos" and sg_platform.lower() != "windows":
+    #         cmd = ["scp", "certs.zip", "{}@{}:/tmp".format(username, node["ip"])]
+    #     else:
+    #         cmd = ["cp", "certs.zip", "/tmp"]
+    #     print(" ".join(cmd))
+    #     proc = Popen(cmd, stdout=PIPE, stderr=PIPE)
+    #     stdout, stderr = proc.communicate()
+    #     if stdout or stderr:
+    #         print(stdout, stderr)
 
 
 def load_cluster_config_json(cluster_config):
--- a/utilities/cluster_config_utils.py
+++ b/utilities/cluster_config_utils.py
@@ -4,7 +4,7 @@ import os
 import re
 
 from keywords.exceptions import ProvisioningError
-from shutil import copyfile, rmtree, make_archive
+from shutil import copyfile, rmtree
 from subprocess import Popen, PIPE
 from distutils.dir_util import copy_tree
 
@@ -82,7 +82,7 @@ def generate_x509_certs(cluster_config, bucket_name, sg_platform):
             if match:
                 username = match.groups()[0].strip()
                 break
-
+    print(username)
     curr_dir = os.getcwd()
     certs_dir = os.path.join(curr_dir, "certs")
     if os.path.exists(certs_dir):
@@ -103,20 +103,7 @@ def generate_x509_certs(cluster_config, bucket_name, sg_platform):
     stdout, stderr = proc.communicate()
     print(stdout, stderr)
 
-    # # zipping the certificates
     os.chdir(curr_dir)
-    # make_archive("certs", "zip", certs_dir)
-
-    # for node in cluster["sync_gateways"]:
-    #     if sg_platform.lower() != "macos" and sg_platform.lower() != "windows":
-    #         cmd = ["scp", "certs.zip", "{}@{}:/tmp".format(username, node["ip"])]
-    #     else:
-    #         cmd = ["cp", "certs.zip", "/tmp"]
-    #     print(" ".join(cmd))
-    #     proc = Popen(cmd, stdout=PIPE, stderr=PIPE)
-    #     stdout, stderr = proc.communicate()
-    #     if stdout or stderr:
-    #         print(stdout, stderr)
 
 
 def load_cluster_config_json(cluster_config):
--- a/utilities/mobile_server_pool.py
+++ b/utilities/mobile_server_pool.py
@@ -170,7 +170,7 @@ def cleanup_for_blocked_nodes(job_name=None):
         job_name = length_array[0]
         build_id = length_array[1]
         log_info("Releasing node {} for job {}".format(doc_id, job_id))
-        if job_name is not None or not is_jenkins_job_running(job_name, build_id) :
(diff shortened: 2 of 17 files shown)


utilities/mobile_server_pool.py
```diff
@@ -115,9 +165,12 @@ def cleanup_for_blocked_nodes():
         job_id = doc["username"]
         if job_id == "":
             job_id = "test_job:123"
-        job_name, build_id = job_id.split(":")
```
mgcouchbase, line 118: this is correct and fast ,we do not need to refactor it.

utilities/mobile_server_pool.py
```diff
@@ -210,8 +278,30 @@ def is_jenkins_job_running(job_name='Net-windows-TestSever-Funtional-tests', bui
...
+                sgw_node_list = opts.sgw_nodes.strip().split(',')
+                for node in sgw_node_list:
+                    pool_json_str += '"{}", '.format(node)
+            if opts.load_balancer_nodes:
+                lp_node_list = opts.load_balancer_nodes.strip().split(',')
+                for node in lp_node_list:
+                    pool_json_str += '"{}", '.format(node)
+            pool_json_str = pool_json_str.rstrip(', ')
+            pool_json_str += "],"
+            if opts.sgw_nodes:
```
mgcouchbase, line 291: I feel,  if we can keep only One if condition for opts.sgw_nodes:  and keep all the related code , so it will be very easy to read. (this way we might need to add two duplicate line )

utilities/mobile_server_pool.py
```diff
@@ -30,6 +55,12 @@ def get_nodes_from_pool_server(num_of_nodes, nodes_os_type, node_os_version, job
     for row in sdk_client.n1ql_query(query):
         doc_id = row["id"]
         is_node_reserved = reserve_node(doc_id, job_name_requesting_node)
+        vm_alive = check_vm_alive(doc_id)
+        if not vm_alive:
+            query_str = "update `{}` set state=\"VM_NOT_ALLIVE\" where meta().id='{}' " \
+                "and state='available'".format(doc_id)
+            query = N1QLQuery(query_str)
+            sdk_client.n1ql_query(query)
```
euniceatcouchbase, line 63: I assume when vm_alive is not true, then this vm will not be added to the pool_list. do you need to add a "continue" operation after this line 63?

utilities/mobile_server_pool.py
```diff
@@ -115,9 +165,12 @@ def cleanup_for_blocked_nodes():
...
         if job_id == "":
             job_id = "test_job:123"
-        job_name, build_id = job_id.split(":")
+        length_array = []
+        length_array = job_id.split(":")
+        job_name = length_array[0]
+        build_id = length_array[1]
         log_info("Releasing node {} for job {}".format(doc_id, job_id))
-        if not is_jenkins_job_running(job_name, build_id):
+        if job_name is not None or not is_jenkins_job_running(job_name, build_id):
```
euniceatcouchbase, line 173: I am ok with this change, just thinking of how to prevent cleaning a node which is still running a jenkins job? any thought?

utilities/cluster_config_utils.py
```diff
@@ -82,7 +82,7 @@ def generate_x509_certs(cluster_config, bucket_name, sg_platform):
             if match:
                 username = match.groups()[0].strip()
                 break
-
+    print(username)
```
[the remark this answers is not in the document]
sridevi-15: I cannot use log_info here, so I had to use print. I need to import utils to use log_info and utils already imports this file and it goes into infinite loop for importing and python does not allow that. That is why in few files you see print if log_info cannot be used

utilities/cluster_config_utils.py
```diff
@@ -103,20 +103,7 @@ def generate_x509_certs(cluster_config, bucket_name, sg_platform):
     stdout, stderr = proc.communicate()
     print(stdout, stderr)
```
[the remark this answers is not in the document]
sridevi-15: same as above, that is why print was used earlier

utilities/mobile_server_pool.py
```diff
@@ -11,6 +14,28 @@
...
+    :param job_name_requesting_node: Name of the job requesting for node
+    :return: list of nodes reserved
+    """
+    # Getting list of available nodes through n1ql query
+    query_str = "select count(*) from `{}` where os='{}' " \
+                "AND os_version='{}' AND state='available'".format(BUCKET_NAME, nodes_os_type, node_os_version)
+    query = N1QLQuery(query_str)
+    for row in sdk_client.n1ql_query(query):
+        count = row["$1"]
+    print(count)
```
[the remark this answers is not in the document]
sridevi-15: printing here to return count. When I get the available count , based on what it printed, I get the server nodes. This print is not for logging, this is give the count while running on jenkins shell.

utilities/mobile_server_pool.py
```diff
@@ -11,6 +14,28 @@
 PASSWORD = 'esabhcuoc'
 BUCKET_NAME = "QE-mobile-pool"
 sdk_client = Bucket("couchbase://{}/{}".format(SERVER_IP, BUCKET_NAME))
+SSH_NUM_RETRIES = 3
```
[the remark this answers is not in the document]
sridevi-15: yeah, the reason is to allign with server retries. Server team while checking for sshing , it has retries for 3 times, so want to go with that.

utilities/mobile_server_pool.py
sridevi-15: this is not refactoring, this is according to new logic while reserving nodes from server team. This current logic is only expecting the name as job:build, but cannot handle if username has more than that.  When we reserver nodes, we should add other names to username and it fails to get job and build name.

utilities/mobile_server_pool.py
sridevi-15: yeah I already tried to do that. Closing quote will be different if u have one if condition . json is not getting created properly

utilities/mobile_server_pool.py
sridevi-15: we clean up in two different ways, 
one is release nodes -> this method used at post build actions once jenkins job is successful
second one is clean up blocked nodes -> This we have to run once in a while to clean up all the blocked nodes to available. if for any reason release nodes (above method) does not release nodes( could be jenkins down, power outages, unexpected job aborted), it leaves the nodes 'blocked' though jenkins jobs not running. Clean_up_blocked_nodes : clean up all blocked nodes if we notice nodes are occupied unnecessarily, but at the same time we should not interrupt if any of the jobs running.

utilities/mobile_server_pool.py
sridevi-15: oh , thanks for noticiing! Good catch.  I made the change.

utilities/mobile_server_pool.py
mgcouchbase: You mean we may get more values separated by colon ?

utilities/mobile_server_pool.py
euniceatcouchbase: make sense, sounds good to me.

PR #1702, opened 2020-05-15
Fix/sgw tagging

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Added oscertify tags and provided tags to run tests as components
- cleaned up overlapping tags and tagged oscertify only one of the parameter for each test


testsuites/syncgateway/functional/tests/test_sync-function-reject.py
```diff
@@ -16,7 +16,8 @@
 @pytest.mark.syncgateway
 @pytest.mark.attachments
 @pytest.mark.session
-@pytest.mark.channel
+@pytest.mark.sync
```
euniceatcouchbase, line 19: just to ensure that you do mean to add the sync mark.

PR #1708, opened 2020-05-29
Fix/cm 435 value serialzer fix for DotNet

mgcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Updated .net value serializer, as per the ios code.
- http://uberjenkins.sc.couchbase.com:8080/view/04%20sanity/job/Net-windows-TestSever-Funtional-tests-sanity2/28/console
-With this update code will match with ios and android

Xcode  change
-  if (value!.hasPrefix("\"") && value!.hasSuffix("\"")) {
            let start = value!.index(value!.startIndex, offsetBy: 1)
            let end = value!.index(value!.endIndex, offsetBy: -1)
            let range = start..<end

Strings tested  with:  ""reter""RETRE"", "reter""RET"RE"


testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
```diff
@@ -665,7 +665,7 @@ def test_non_blocking_custom_conflicts_resolution(params_from_base_test_setup, r
             new_docs_body[doc_id] = [data]
             data = property_updater(data)
             random_value = random_string(length=10, printable=True)
-            data["update_during_CCR"] = random_value
+            data["update_during_CCR"] = random_value + '"'
```
euniceatcouchbase, line 668: is this for your testing only? I don't see a reason to append a double quote to the end of the random value of string.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
```diff
@@ -695,10 +695,13 @@ def test_non_blocking_custom_conflicts_resolution(params_from_base_test_setup, r
                                                                           "with delayed local win"
             assert "random" not in cbl_doc, "CCR failed to resolve conflict with delayed local win"
             assert "cbl_random" not in sg_doc, "CCR failed to resolve conflict with delayed local win"
-            assert new_docs_body[doc_id][1]["update_during_CCR"] == cbl_doc["update_during_CCR"], "CCR failed to " \
-                                                                                                  "resolve conflict " \
-                                                                                                  "with delayed " \
-                                                                                                  "local win"
+            log_info(cbl_doc["update_during_CCR"] + "resolve conflict with delayed local win" + new_docs_body[doc_id][1][
+                "update_during_CCR"])
+            assert new_docs_body[doc_id][1]["update_during_CCR"] == cbl_doc[
```
euniceatcouchbase, line 700: seems your branch doesn't have the latest master code. The latest master has the replace('"', '') code you added earlier. check this: https://github.com/couchbaselabs/mobile-testkit/blame/master/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py#L700

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/ValueSerializer.cs
```diff
@@ -117,7 +117,8 @@ internal static object DeserializeValue(String value)
             }
             else if (value.StartsWith("\"") && value.EndsWith("\""))
             {
-                return value.Replace("\"", String.Empty);
+                string st = value.Remove(0, 1);
+                return st.Remove(st.LastIndexOf("\""));
```
euniceatcouchbase, line 121: should line 121 be return st.Remove(st.LastIndexOf("\""), 1) ? say you have a string value \"couchbase\", st will be "couchbase\", and line 121 will give you "couchbase, the last quote is completed removed, not only the back slash \

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
mgcouchbase: removed ,used it to make sure all aways random string has cote.

CBLClient/Apps/CBLTestServer-Dotnet/TestServer/ValueSerializer.cs
mgcouchbase: Like python if just add """ I am seeing errors , so we need to keep the back slash.   Goal here is to remove the cotes around the string . Line 120 removes the starting cote and 121 remove the ending cote. I am use the trim methods, but it will remove all the ending cotes . We want to remove only one ending cote, hence I needed to go with this approach

mgcouchbase: Added the swift code for this change.

PR #1715, opened 2020-06-12
optimize blob code use the common util function

euniceatcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- now a common function in utils is available to fill up embedded resource path by platform, switch to use it to avoid incorrect file path is specified in test cases


sridevi-15:
this is much cleaner one

euniceatcouchbase:
> It works for IOS and Dotnet too?
yes, pls check https://github.com/couchbaselabs/mobile-testkit/blob/master/keywords/utils.py#L431, you need to pass platform name, the function will return you a proper resource location.

euniceatcouchbase:
> this is much cleaner one

the util function was written based on your review comments last time.

PR #1721, opened 2020-07-16
fix index issues

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- added logic for retries to check all index deleted when buckets are deleted. This will help in wait until all index cleaned up before recreating same buckets

http://uberjenkins.sc.couchbase.com:8080/job/cen7-sync-gateway-functional-tests-topology-cc/182/testReport/

### DIFF
--- a/keywords/couchbaseserver.py
+++ b/keywords/couchbaseserver.py
@@ -184,6 +184,18 @@ class CouchbaseServer:
         if len(bucket_names) != 0:
             raise CBServerError("Failed to delete all of the server buckets!")
 
+        # verify all indexes are deleted
+        count = 0
+        index_url = self.url.replace("8091", "9102")
+        while count < 5:
+            resp = self._session.get("{}/getIndexStatus".format(index_url))
+            resp_obj = resp.json()
+            if "status" not in resp_obj:
+                break
+            else:
+                count += 1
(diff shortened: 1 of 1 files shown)


keywords/couchbaseserver.py
```diff
@@ -184,6 +184,18 @@ def delete_buckets(self):
         if len(bucket_names) != 0:
             raise CBServerError("Failed to delete all of the server buckets!")
 
+        # verify all indexes are deleted
+        count = 0
+        index_url = self.url.replace("8091", "9102")
+        while count < 5:
+            resp = self._session.get("{}/getIndexStatus".format(index_url))
```
mgcouchbase, line 191: When I was adding retry I have explore this api  "https://pypi.org/project/retry/"  and it is very good. I couldn't use it since my use case are more than a api.

euniceatcouchbase: sounds great. by using this retry api, do we need to add this to the requirement lists to ensure the python package gets installed?

sridevi-15: cool, yes it needs new package. Let's try this out when we do it next time

PR #1736, opened 2020-09-02
Fix/magma support

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- added support to add magma
- updated conftest files to run with magma to all kinds of tests

Jenkins jobs : Full set of sgw regression ran 
http://uberjenkins.sc.couchbase.com:8080/view/04%20sync-gateway/job/_sync-gateway-upstream/1421/
jobs failed got reran with magma enabled and they are successful

### DIFF
--- a/testsuites/syncgateway/functional/tests/conftest.py
+++ b/testsuites/syncgateway/functional/tests/conftest.py
@@ -144,6 +144,10 @@ def pytest_addoption(parser):
                      action="store_true",
                      help="delta-sync: Enable delta-sync for sync gateway")
 
+    parser.addoption("--magma-storage",
+                     action="store_true",
+                     help="magma-storage: Enable magma storage on couchbase server")
+
     parser.addoption("--cbs-ce", action="store_true",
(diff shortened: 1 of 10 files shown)


testsuites/syncgateway/functional/tests/conftest.py
```diff
@@ -144,6 +144,10 @@ def pytest_addoption(parser):
                      action="store_true",
                      help="delta-sync: Enable delta-sync for sync gateway")
 
+    parser.addoption("--magma-storage",
```
euniceatcouchbase, line 147: this set default value to true. what if for non-magma version, will setting magma-storage to true break anything if server version is not magma? if non-magma version is with it, then nothing need to change. if there is something not compatible, then setting default value to false may be easier

euniceatcouchbase (approved):
have one comment about default value, otherwise they all look good

mgcouchbase (approved):

PR #1744, opened 2020-09-15
Fix/cm 470 log file collect failure

euniceatcouchbase:
#### Fixes #. sometimes logs are not generated, open log file will cause post run failure. the fix will prevent test fail twice

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- added file exist condition check before opening.


sridevi-15 (approved):
Hope you ran the test locally. As it is just if and else condition , it does not really require jekins job as logic hasn't change much

mgcouchbase (approved):

PR #1747, opened 2020-09-17
regression test fixes

mgcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- 
-

### DIFF
--- a/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
+++ b/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
@@ -512,9 +512,12 @@ def test_push_replication_for_20mb_doc(params_from_base_test_setup, attachment_g
     cbl_db = params_from_base_test_setup["source_db"]
     sync_gateway_version = params_from_base_test_setup["sync_gateway_version"]
     liteserv_platform = params_from_base_test_setup["liteserv_platform"]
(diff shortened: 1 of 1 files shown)


sridevi-15 (commented):
if we are skipping this test on Android and iOS with device enabled, this test won't be running at all. Is there any alternative way to cover this scenario? If not we can discuss with the dev team and ask them  if they can help so that we can have this coverage

mgcouchbase:
Looks like this test is added to replicate the blob use case, Now we have separate test to cover blob usecase. Our test server app is converting this image into more than 2100.MB docs which is making the device to crash. This one I debugged with Jay and as per him we have to move this test for stress testing to check how many max docs we support on device.

euniceatcouchbase:
by some reason I don't think we have blob specific test cases written except the blob constructor test I added early this year. This test case failed on java also. probably this is time to review the test case with dev, to understand how large attachment is handled, if the proper way of handling it changed to blob, instead of skipping it, then we re-implement it in a correct approach using blob.

mgcouchbase:
As per Jay, we need to find out the higher limit for cbl log size.

euniceatcouchbase (commented):
I agree with Sridevi, seems Manasa already has discussed with Jay and got some root cause. Therefore we probably need to re-write this test case in a different way, (such as use blob) in order to fix the issue. we may need to add a flag to create_bulk_docs if use blob, if so, then make the attachments as blob.

PR #1749, opened 2020-09-18
Custom rerun failed tests report & added Merge support

mgcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

-  As xml parser merged few days back, adding report xml report merge for --rerun-tests
-- Adding merge xmls option and it is used when list of xmls are provided

Example run:  http://uberjenkins.sc.couchbase.com:8080/view/testkit-testing/job/project100-exclude-include-topo/24/


utilities/xml_parser.py
```diff
@@ -0,0 +1,170 @@
...
+        doc = xml.dom.minidom.parse(xml_file)
+        testresultelem = doc.getElementsByTagName("testsuites")
+        testsuitelem = testresultelem[0].getElementsByTagName("testsuite")
+        for ts in testsuitelem:
+            testcaseelem = ts.getElementsByTagName("testcase")
+            for tc in testcaseelem:
+                tcname = tc.getAttribute("name")
+                if tc.getElementsByTagName("failure") or tc.getElementsByTagName("error"):
+                    failed = True
+                    failed_tests.append(tcname.split('[')[0])
```
sridevi-15, line 82: remove the split for '['. We need to rerun with parameters that are failed, but not all

utilities/xml_parser.py
```diff
@@ -0,0 +1,170 @@
...
+                else:
+                    failed = False
+                    passed_tests.append(tcname)
+
+    if failed_tests:
```
sridevi-15, line 87: is this condition required as this is always be true. Though it is empty, it is still treated as true. May be you want to add the condtion that length of failed tests is more than 0

utilities/xml_parser.py
```diff
@@ -0,0 +1,170 @@
...
+        failed_tests = transform_and_write_to_file(failed_tests, "failed_tests.conf")
+
+    if passed_tests:
```
sridevi-15, line 90: same as above

sridevi-15 (changes requested):
changes looks good, minor changes. Btw, all flake8 issues fixed, right? when I used xml parser in my local machine, it is having flake8 errors.
Also to make sure this works, can you provide jenkins successful run?

utilities/xml_parser.py
```diff
@@ -0,0 +1,170 @@
...
+
+if __name__ == "__main__":
+    usage = """usage: xml_parser.py
+       usage: python mobile_server_pool.py
+       --file=xml-file path
+       """
+    parser = OptionParser(usage=usage)
+
+    parser.add_option("--xml-file",
+                      action="store", dest="xml_file", default="http://uberjenkins.sc.couchbase.com:8080/job/cen7-sync-gateway-functional-tests-sgreplicate2/lastCompletedBuild/artifact/results/results.xml",
```
euniceatcouchbase, line 162: how about not setting default? or default location without job name, such as a regex pattern so each job can fill up with its name? just curious.

euniceatcouchbase (approved):

sridevi-15:
@mgcouchbase , alreaddy reviewed last time itself, I think I requested for jenkins job which u provided now. Can you check jenkins results?
#1 and #3 jenkins url doing same, rerunning passed tests and merging the rerun url . one thing I noticed is that number of tests shows 47 total where as actual run shows 27. Looks like it is miscalculating . 
 I think what would be ideal is that have #1 to rerun with passed tests  and #3 with failed tests.  - that will help you to test your changes, #1 should still rerun with passed tests and should have failed tests in the results
#3 -  it should pass all the failed with rerun . Not sure why it shows failing in your jenkins run - in this case it should show all 27 tests passed, but only 4 tests ran. 

let me know if u want to sync up to test ur changes 

#2 is just a rerun passed or failed test which I guess u added for xml parser, but not for merging xml which is fine.

sridevi-15 (approved):

PR #1758, opened 2020-10-09
new tests for sgw 2.7.4

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- 4 new tests written to cover delta sync tests which does not have deltas when there are updates on sgw or cbl
one of the test is expected to fail when you run with 2.7.3 or below as it is a bug


euniceatcouchbase (approved):
looks good to me

mgcouchbase (approved):

PR #1763, opened 2020-10-15
Fix for 20 gb test fix for X-android

mgcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- 
-

### DIFF
--- a/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
+++ b/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
@@ -513,7 +513,7 @@ def test_push_replication_for_20mb_doc(params_from_base_test_setup, attachment_g
     sync_gateway_version = params_from_base_test_setup["sync_gateway_version"]
     liteserv_platform = params_from_base_test_setup["liteserv_platform"]
 
-    if sync_gateway_version < "2.5.0" or liteserv_platform == "android" or ("load_from_data_dir" in str(attachment_generator) and liteserv_platform == "xamarin-android"):
+    if sync_gateway_version < "2.5.0" or ("load_from_data_dir" in str(attachment_generator) and (liteserv_platform == "xamarin-android" or liteserv_platform == "android")):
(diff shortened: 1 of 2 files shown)


testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
```diff
@@ -513,7 +513,7 @@ def test_push_replication_for_20mb_doc(params_from_base_test_setup, attachment_g
     sync_gateway_version = params_from_base_test_setup["sync_gateway_version"]
     liteserv_platform = params_from_base_test_setup["liteserv_platform"]
 
-    if sync_gateway_version < "2.5.0" or liteserv_platform == "android":
+    if sync_gateway_version < "2.5.0" or liteserv_platform == "android" or ("load_from_data_dir" in str(attachment_generator) and liteserv_platform == "xamarin-android"):
```
sridevi-15, line 516: this looks like same condition which required for android too. Can you add it to android?

sridevi-15 (changes requested):

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py
mgcouchbase: Eunice , Can you please confirm this

euniceatcouchbase: yes, currently 20mb tests are skipped on android. java fails on it too. we need to find a way to fix it, instead of skipping it

mgcouchbase: Earlier code was skipping all 3 tests , with this change only 1 test will be skipped.  Currently 3 tests are passing in ios and .net (except the android). As you mentioned earlier android is not supported

sridevi-15 (approved):

PR #1764, opened 2020-10-19
cm-472 fix error message evaluation for javaws

euniceatcouchbase:
#### Fixes #. cm-472 java webservice error message handling

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- java webservice error message can be checked for specific exception cause now.
- make javaws and ios the same assertion criteria

### DIFF
--- a/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication.py
+++ b/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication.py
@@ -1343,8 +1343,10 @@ def test_replication_wrong_blip(params_from_base_test_setup):
     with pytest.raises(Exception) as ex:
         replicator.configure(cbl_db, sg_blip_url, continuous=True, channels=channels, replicator_authenticator=replicator_authenticator)
     ex_data = str(ex.value)
-    if liteserv_platform == "ios":
-        assert "Invalid scheme for URLEndpoint url (ht2tp" in ex_data
+
+    if liteserv_platform == "ios" or liteserv_platform.startswith("javaws-"):
+        assert "Invalid scheme for URLEndpoint url" in ex_data
+        assert "ht2tp" in ex_data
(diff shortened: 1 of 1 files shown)


testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication.py
```diff
@@ -1343,8 +1343,10 @@ def test_replication_wrong_blip(params_from_base_test_setup):
     with pytest.raises(Exception) as ex:
         replicator.configure(cbl_db, sg_blip_url, continuous=True, channels=channels, replicator_authenticator=replicator_authenticator)
     ex_data = str(ex.value)
-    if liteserv_platform == "ios":
-        assert "Invalid scheme for URLEndpoint url (ht2tp" in ex_data
+
+    if liteserv_platform == "ios" or liteserv_platform.startswith("javaws-"):
+        assert "Invalid scheme for URLEndpoint url" in ex_data
```
mgcouchbase, line 1348: Can you add TODO here , when dev refactor old error messages to common on all CBLs then we can easily search.

mgcouchbase: Rest looks good

sridevi-15 (approved):

mgcouchbase (approved):

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication.py
euniceatcouchbase: I agree error messages from all platforms should be same. But for this particular issue, android and java desktop doesn't throw proper error message because 3rd party nanohttpd package we used hide the message. I don't know what reason for dotnet. I can file a ticket for dotnet, but we don't have priority to investigate at least for now. Therefore TODO doesn't apply for this test case, maybe for other test cases, I know there are quite many places we have to check per platform.

PR #1791, opened 2021-01-14
Fix/test predictive queries

mgcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- http://uberjenkins.sc.couchbase.com:8080/job/CBLITE_DOTNET-Net-windows-TestSever-Functional-XATTRS-tests/63/
-http://uberjenkins.sc.couchbase.com:8080/job/CBLITE_DOTNET-Net-windows-TestSever-Functional-XATTRS-tests/62/

### DIFF
--- a/keywords/utils.py
+++ b/keywords/utils.py
@@ -83,7 +83,7 @@ def version_and_build(full_version):
         return version_parts[0], None
 
 
-def  ohost_for_url(url):
+def host_for_url(url):
     """ Takes a url in the form of http://192.168.33.10:4985
     and returns an host in the form 192.168.33.10
     """
@@ -345,7 +345,7 @@ def compare_generic_types(object1, object2, isPredictiveResult=False):
         return object1 == object2
     elif isinstance(object1, float) and isinstance(object2, float):
         if isPredictiveResult:
-            return object1 - object2 < 0.2
+            return (object1 - object2) < 0.2
         else:
             return object1 == object2
     elif isinstance(object1, float) and isinstance(object2, int):
--- a/keywords/utils.py
+++ b/keywords/utils.py
@@ -345,7 +345,7 @@ def compare_generic_types(object1, object2, isPredictiveResult=False):
         return object1 == object2
     elif isinstance(object1, float) and isinstance(object2, float):
         if isPredictiveResult:
-            return (object1 - object2) < 0.001
+            return (object1 - object2) < 0.1
         else:
             return object1 == object2
     elif isinstance(object1, float) and isinstance(object2, int):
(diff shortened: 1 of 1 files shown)


keywords/utils.py
```diff
@@ -344,7 +344,10 @@ def compare_generic_types(object1, object2, isPredictiveResult=False):
     elif isinstance(object1, int) and isinstance(object2, int):
         return object1 == object2
     elif isinstance(object1, float) and isinstance(object2, float):
-        return object1 == object2
+        if isPredictiveResult:
+            return (object1 - object2) < 0.1
```
euniceatcouchbase, line 348: you need either use abs() or to ensure the greater one minus the smaller one < 0.1, otherwise, if 1.5 - 10.5 will also meet the criteria of < 0.1.

euniceatcouchbase (changes requested):

keywords/utils.py
pasin: @euniceatcouchbase is right. Also 0.1 precision is a bit big and you could lower that to something like 0.001.

euniceatcouchbase: @pasin I think we have seen the delta between 2 float numbers is 0.01, as predictive query, do we consider it's an issue? if so, we will file a bug

euniceatcouchbase (approved):

keywords/utils.py
```diff
@@ -344,7 +344,10 @@ def compare_generic_types(object1, object2, isPredictiveResult=False):
     elif isinstance(object1, int) and isinstance(object2, int):
         return object1 == object2
     elif isinstance(object1, float) and isinstance(object2, float):
-        return object1 == object2
+        if isPredictiveResult:
```
mgcouchbase, line 347: please wait  I need to test it more

keywords/utils.py
mgcouchbase: mismatch 91041445404867.81 and 91041445404867.8

pasin: Actually, after thinking about this again, fixing the tolerance precision is not ideal as the delta will depend on the number itself. You can use 0.1 as I don't think the 0.1 mismatch is the issue.

pasin: I have found some discussion (a proposal) on Swift's forum to provide a method to compare float. I haven't thoroughly read about the proposal and comment yet but I understand that the proposal will try to come up with a reasonable tolerance precision based on the number using the info such as ulpOfOne.

You could review the proposal and adopt the implementation (could be cross-platform) for comparing float.

https://github.com/apple/swift-evolution/blob/master/proposals/0259-approximately-equal.md
https://forums.swift.org/t/se-0259-approximate-equality-for-floating-point

PR #1808, opened 2021-02-24
Fix/recoverytype

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Added catching exception and retry the rest call 
 jenkins job : http://uberjenkins.sc.couchbase.com:8080/job/cen7-sync-gateway-tests-sanity/403/

### DIFF
--- a/keywords/couchbaseserver.py
+++ b/keywords/couchbaseserver.py
@@ -742,11 +742,14 @@ class CouchbaseServer:
         count = 0
         max_retries = 5
         while count < max_retries:
-            resp = self._session.post(
-                "{}/controller/setRecoveryType".format(self.url),
-                headers={"Content-Type": "application/x-www-form-urlencoded"},
-                data=data
-            )
+            try:
+                resp = self._session.post(
+                    "{}/controller/setRecoveryType".format(self.url),
+                    headers={"Content-Type": "application/x-www-form-urlencoded"},
+                    data=data
+                )
+            except HTTPError:
+                log_info("Got http error while trying to recover the server, so trying one more time")
             if resp.status_code == 200:
                 break
             count += 1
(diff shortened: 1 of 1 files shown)


keywords/couchbaseserver.py
```diff
@@ -741,11 +741,14 @@ def recover(self, server_to_recover):
...
-                data=data
-            )
+            try:
+                resp = self._session.post(
+                    "{}/controller/setRecoveryType".format(self.url),
+                    headers={"Content-Type": "application/x-www-form-urlencoded"},
+                    data=data
+                )
+            except HTTPError:
+                log_info("Got http error while trying to recover the server, so trying one more time")
```
mgcouchbase, line 751: you do not want to raise the exception and fail the test ?

euniceatcouchbase: what's the difference between HTTPError and resp.status_code not equal to 200? I agree with the original code, should not raise exception here because retries are expected.

euniceatcouchbase (approved):

keywords/couchbaseserver.py
sridevi-15: @mgcouchbase .No we should not raise the exception. The idea is to retry . We get http error  which is expected and have retry until it is successful

@euniceatcouchbase . The above also does the same thing what you have mentioned. If http error raised, it will continue and count get increase and it is not status code == 200, so it won't break the loop. Wondering if there is any difference

euniceatcouchbase: got it, I am fine with the implementation. thumb up.

mgcouchbase (approved):

PR #1852, opened 2021-05-27
support install key function to non-{pool.json} defined machines

euniceatcouchbase:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- extended the option to allow specify host ips then install key on those

### DIFF
--- a/libraries/utilities/install_keys.py
+++ b/libraries/utilities/install_keys.py
@@ -1,14 +1,16 @@
 import os
 import sys
+import json
 
 from optparse import OptionParser
 from generate_clusters_from_pool import get_hosts
 import paramiko
 
 
-def install_keys(public_key_path, user_name, ssh_password):
+def install_keys(public_key_path, user_name, ssh_password, hosts):
 
-    hosts, _ = get_hosts()
+    if hosts is None:
+        hosts, _ = get_hosts()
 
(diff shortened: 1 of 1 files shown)


libraries/utilities/install_keys.py
```diff
@@ -1,14 +1,16 @@
...
 import sys
+import json
 
 from optparse import OptionParser
 from generate_clusters_from_pool import get_hosts
 import paramiko
 
 
-def install_keys(public_key_path, user_name, ssh_password):
+def install_keys(public_key_path, user_name, ssh_password, hosts):
```
sridevi-15, line 10: Can you have hosts with default value as 'None'
install_keys(public_key_path, user_name, ssh_password, hosts=None): 
Just in case if this method used any where in the future, it won't force to use hosts field

mgcouchbase (approved):

sridevi-15 (changes requested):
and also for my clarification, you extended this , so that we can install keys on the ips listed on command line and ignore ips on pool.json?

libraries/utilities/install_keys.py
euniceatcouchbase: fixed, thanks for reminding

PR #1859, opened 2021-06-10
updated the test to have sgw fail when disabled xattrs with xattrs key enabled

sridevi-15:
…y enabled

#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- separate out the test as it is different behavior from other tests

### DIFF
--- a/testsuites/syncgateway/functional/tests/test_sgw_config.py
+++ b/testsuites/syncgateway/functional/tests/test_sgw_config.py
@@ -245,6 +245,7 @@ def test_invalid_external_jspath(params_from_base_test_setup, setup_jsserver):
     cluster_helper = ClusterKeywords(cluster_config)
     cluster_hosts = cluster_helper.get_cluster_topology(cluster_config)
     sg_url = cluster_hosts["sync_gateways"][0]["public"]
+    flag = False
 
     if sync_gateway_version < "3.0.0":
         pytest.skip("this feature not available below 3.0.0")
@@ -268,10 +269,13 @@ def test_invalid_external_jspath(params_from_base_test_setup, setup_jsserver):
 
     try:
         requests.get(sg_url, timeout=30)
-        assert False, "Sync gateway started successfully with invalid external jsfile "
+        flag = True
     except Exception as he:
         log_info(str(he))
         log_info("Expected to have sync gateway fail to start")
+
+    if flag:
+        assert False, "Sync gateway started successfully with invalid external jsfile"
 
 
 @pytest.mark.syncgateway
--- a/testsuites/syncgateway/functional/tests/test_sgw_config.py
+++ b/testsuites/syncgateway/functional/tests/test_sgw_config.py
@@ -3,7 +3,7 @@ import time
 import os
 import requests
 import subprocess
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, ConnectionError
 
 from keywords.utils import log_info, get_local_ip
 from keywords.ClusterKeywords import ClusterKeywords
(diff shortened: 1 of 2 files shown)


testsuites/syncgateway/functional/tests/test_xattrs_grants.py
```diff
@@ -690,6 +691,56 @@ def test_missing_xattrs_key(params_from_base_test_setup, missing_type):
...
+
+    if sync_gateway_version < "3.0.0" or xattrs_enabled:
+        pytest.skip('SGW version is not 3.0 and above or xattrs has to be disabled')
+
+    user_custom_channel1 = "channel1"
+    sg_conf_name = "custom_sync/sync_gateway_custom_sync"
+    sg_config = sync_gateway_config_path_for_mode(sg_conf_name, mode)
+
+    temp_sg_config, _ = copy_sgconf_to_temp(sg_config, mode)
+    temp_sg_config = replace_string_on_sgw_config(temp_sg_config, "{{ user_xattrs_key }}", "")
```
euniceatcouchbase, line 720: does this enabling user_xattrs_key? does it need a string or true value?

sridevi-15: string value

testsuites/syncgateway/functional/tests/test_sgw_config.py
```diff
@@ -268,11 +269,14 @@ def test_invalid_external_jspath(params_from_base_test_setup, setup_jsserver):
 
     try:
         requests.get(sg_url, timeout=30)
-        assert False, "Sync gateway started successfully with invalid external jsfile "
```
mgcouchbase, line 271: What is advantage of making this change.

testsuites/syncgateway/functional/tests/test_xattrs_grants.py
```diff
@@ -690,6 +691,56 @@ def test_missing_xattrs_key(params_from_base_test_setup, missing_type):
...
+        sgrestart = mp.submit(sg.restart, temp_sg_config, cluster_config)
+        time.sleep(60)
+    sgrestart.result(timeout=80)
+
+    try:
+        requests.get(sg_url, timeout=30)
+        flag = True
+    except Exception as he:
+        log_info(str(he))
+        log_info("Expected to have sync gateway fail to start")
```
mgcouchbase, line 738: You can validate the exception string, by adding the assert in the catch block.

euniceatcouchbase (approved):

testsuites/syncgateway/functional/tests/test_sgw_config.py
sridevi-15: if the assert Fails, it is caught in exception rathar than showing that assertion is failed. With this change, this will catch only the exception with request.get, but not for assert False

testsuites/syncgateway/functional/tests/test_xattrs_grants.py
sridevi-15: even if u validate, you have to say "assert True" which does not require. If Test is passed, there is no explicit requirement for assertion True.

mgcouchbase: I mean, if we assert the exception string that we will cover the exception validation as well

mgcouchbase (approved):

testsuites/syncgateway/functional/tests/test_xattrs_grants.py
sridevi-15: okay, got it. I made the change that can catch right exception and remove the addition flag variable

mgcouchbase (commented):
Looks good now.

PR #1906, opened 2022-01-18
Fix/c ios fixes

sridevi-15:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- have c-ios fixes

### DIFF
--- a/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
+++ b/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
@@ -1,4 +1,5 @@
 import pytest
+import time
 
 from time import sleep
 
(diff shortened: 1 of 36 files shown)


testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
```diff
@@ -682,7 +683,9 @@ def test_non_blocking_custom_conflicts_resolution(params_from_base_test_setup, r
 
     sg_docs_content = sg_client.get_bulk_docs(sg_url, sg_db, doc_ids, session)[0]
     cbl_docs = db.getDocuments(cbl_db, doc_ids)
-
+    import time
```
euniceatcouchbase, line 686: could you please move the import to the beginning of the file? in-place import will be hard to maintain for future.

euniceatcouchbase (changes requested):
one minor change required.

testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_custom_conflicts.py
sridevi-15: Sure, The reason I left it there inside the test is that it will import time only when test run, otherwise it can avoid importing which is unnecessary for other tests. Not sure whether it is advantage to have it inside the side if only single test required. To be consistent with all remaining tests, I moved to the top of the test file

euniceatcouchbase (approved):

PR #1930, opened 2022-09-01
Adjust the ud tags check to ignore binary files

gilkal:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- 
-


hthomps-cb (approved):
After discussion with developers about the issue, this fix appears to be adequate

PR #1937, opened 2022-09-12
Fix/server witl ssl

gilkal:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- 
-


hthomps-cb (changes requested):
There's a couple of typos in line 33: 'overriting' -> 'overwriting' and 'isable_tls' -> 'disable_tls'

hthomps-cb (approved):

gilkal:
> There's a couple of typos in line 33: 'overriting' -> 'overwriting' and 'isable_tls' -> 'disable_tls'.
> 
> Also it's pedantic but the `is True` in line 30 is redundant

Thanks, I've fixed the typos.

Is, I like adding the explicit value, for a slightly better readability, but that's just me. I've removed it as well.

PR #1942, opened 2022-09-20
Increase sdk timeout 4

saigon:
#### Fixes #.

- [ ] Ran `flake8`
- [ ] Ran `run_repo_tests.sh`

- 
-


testsuites/CBLTester/CBL_Functional_tests/conftest.py
```diff
@@ -717,10 +743,12 @@ def params_from_base_test_setup(request, params_from_base_suite_setup):
 
     if not use_local_testserver and create_db_per_test:
         log_info("Starting TestServer...")
+        testserver.stop()
         if device_enabled:
             testserver.start_device(log_filename)
         else:
             testserver.start(log_filename)
+        time.sleep(2)
```
gilkal, line 751: Is there a way we can avoid sleep?

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
```diff
@@ -0,0 +1,44 @@
...
+        ''' Create scope with different name '''
+        self.server_obj.create_scope(scope_name)
+
+    def test_create_collection(self, scope_name, collection_name):
+        '''
+        Test create collection
+        '''
+        self.server_obj.create_collection()
+
+    def test_create_scope_already_exist(self, scope_name):
```
gilkal, line 19: I wonder if we need that test - maybe the developers have the same test running in their suite?

gilkal, line 19: Maybe it's worth to have a cleanup at the end of the test, deleting the scope? It's generally a good practice, although maybe we have a different way to cleanup at the end of the test?

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
```diff
@@ -0,0 +1,44 @@
...
+            else:
+                raise("Server could create another scope with same name")
+        
+    def test_remove_scope(self, scope_name):
+        '''
+        @summary: Testing remove scope
+        '''
+        self.server_obj.remove(scope_name, collection_name)
+
+    def test_remove_collection(self, scope_name, collection_name):
```
gilkal, line 39: Does this test depend on one of the tests before? If it is, it might be worth to make it independent, because dependencies between tests can potentially cause problems.

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
```diff
@@ -0,0 +1,44 @@
...
+        try:
+            self.server_obj.create_scope(scope_name)
+        except Exception as e:
+            if "Fail" in str(e):
+                print("Test failed as expected")
+                pass
+            else:
+                raise("Server could create another scope with same name")
+        
+    def test_remove_scope(self, scope_name):
```
gilkal, line 33: Does this test depend on one of the tests before? If it is, it might be worth to make it independent, because dependencies between tests can potentially cause problems.

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
saigon: This test to ensure that data in current scope or collection will not be deleted if we create a new ones with same name

saigon: This test to ensure that data in current collection will not be deleted if we create a new one with same name

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
saigon: This test is independent test

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
saigon: This test also independent test.  It will remove scope in database

testsuites/CBLTester/CBL_Functional_tests/APITests_sc/test_basic_sc.py
gilkal: Do the developers have that test in their test suite? If they do, maybe it's not worth adding it, because it takes time to run with the whole setup, etc.?

testsuites/CBLTester/CBL_Functional_tests/conftest.py
gilkal: The thing about sleeps is that they might change between different machines: 2 might work fro one machine but not for the other, which might cause the test to be flaky. Is there a way to query that the CBL has started, rather than wait?

gilkal (approved):
It's n important commit, so I'm approving it. It's worth nothing my comments about the usage of sleep and test cleanup.

PR #1954, opened 2022-10-10
Feature/more sc tests

gilkal:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- 
-

### DIFF
--- a/testsuites/syncgateway/functional/tests/conftest.py
+++ b/testsuites/syncgateway/functional/tests/conftest.py
@@ -535,10 +535,10 @@ def params_from_base_suite_setup(request):
     # clean up firewall rules if any ports blocked for server ssl testing
     clear_firewall_rules(cluster_config)
     # Stop all sync_gateway and sg_accels as test finished
-    #c = cluster.Cluster(cluster_config)
+    c = cluster.Cluster(cluster_config)
 
     # Comment this out to use skip-provisions
-    #c.stop_sg_and_accel()
+    c.stop_sg_and_accel()
 
     # Delete png files under resources/data
     clear_resources_pngs()
--- a/testsuites/syncgateway/functional/tests/conftest.py
+++ b/testsuites/syncgateway/functional/tests/conftest.py
@@ -535,10 +535,10 @@ def params_from_base_suite_setup(request):
     # clean up firewall rules if any ports blocked for server ssl testing
     clear_firewall_rules(cluster_config)
     # Stop all sync_gateway and sg_accels as test finished
-    c = cluster.Cluster(cluster_config)
+    #c = cluster.Cluster(cluster_config)
 
     # Comment this out to use skip-provisions
-    c.stop_sg_and_accel()
+    #c.stop_sg_and_accel()
 
     # Delete png files under resources/data
     clear_resources_pngs()
--- a/testsuites/syncgateway/functional/tests/test_cbs_collections.py
+++ b/testsuites/syncgateway/functional/tests/test_cbs_collections.py
@@ -68,7 +68,7 @@ def scopes_collections_tests_fixture(params_from_base_test_setup):
         # Create a user
         pre_test_user_exists = admin_client.does_user_exist(db, sg_username)
         if pre_test_user_exists is False:
-            sg_client.create_user(sg_admin_url, db, sg_username, sg_password, channels, auth)
+            sg_client.create_user(sg_admin_url, db, sg_username, sg_password, auth=auth)
 
         # Create a SGW session
         cookie, session_id = sg_client.create_session(sg_admin_url, db, sg_username, auth=auth)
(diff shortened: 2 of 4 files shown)


testsuites/syncgateway/functional/tests/conftest.py
```diff
@@ -535,10 +535,10 @@ def params_from_base_suite_setup(request):
     # clean up firewall rules if any ports blocked for server ssl testing
     clear_firewall_rules(cluster_config)
     # Stop all sync_gateway and sg_accels as test finished
-    c = cluster.Cluster(cluster_config)
+    #c = cluster.Cluster(cluster_config)
```
abhayayay, line 538: I think making these changes to your local is good to run --skip-provisioning. 
You might want to reconsider merging it to the master.

testsuites/syncgateway/functional/tests/conftest.py
```diff
@@ -535,10 +535,10 @@ def params_from_base_suite_setup(request):
...
 
     # Comment this out to use skip-provisions
-    c.stop_sg_and_accel()
+    #c.stop_sg_and_accel()
```
abhayayay, line 541: I think making these changes to your local is good to run --skip-provisioning.
You might want to reconsider merging it to the master.

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -6,27 +6,32 @@
...
 def scopes_collections_tests_fixture(params_from_base_test_setup):
     try:  # To be able to teardon in case of a setup error
         # get/set the parameters
-        session_id = None
-        pre_test_db_exists = None
-        pre_test_user_exists = None
+        global admin_client
+        global cb_server
+        global sg_username
+        global channels
```
abhayayay, line 33: should we really be using global variables? Yield should be enough to get these variables from the fixture.

gilkal: Thanks. Yes, I agree that global variables is not ideal.  I could yield them from the setup, but that would mean a lot of variables coming out, and some might not be used. Balancing between these considerations and making the global (which is a bad practice, I agree), only in the test file, I think I'm slightly preferring them as global. Other views are also valid, of course.

testsuites/syncgateway/functional/tests/conftest.py
gilkal: Sorry, that shouldn't be there, thanks for spotting it!

testsuites/syncgateway/functional/tests/conftest.py
gilkal: Sorry, that shouldn't be there, thanks for spotting it!

hthomps-cb (commented):
Line 231: This tests if the string is not null and the list is not null, but we want to check if the string is a prefix for one of the IDs in the list, so you want to change this to a loop that tests the prefix against the ids in the list

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -109,11 +112,138 @@ def test_document_only_under_named_scope(scopes_collections_tests_fixture, teard
...
+    shared_found_user_2 = False
+
+    # 4. Check that the users can only see the documents in their channels
+    for doc in user_1_docs_ids:
+        if user_2_doc_prefix in doc:
+            pytest.fail("A document is available in a channel that it was not assigned to. Document prefix: " + user_2_doc_prefix + ". The document: " + doc)
+        if shared_doc_prefix in doc:
+            shared_found_user_1 = True
+        if doc not in wildcard_user_docs_ids:
+            pytest.fail("The document " + doc + " was not accessiable even though the user was given all documents access")
```
hthomps-cb, line 220: typo: accessiable -> accessible

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -109,11 +112,138 @@ def test_document_only_under_named_scope(scopes_collections_tests_fixture, teard
...
+    for doc in user_2_docs_ids:
+        if user_1_doc_prefix in doc:
+            pytest.fail("A document is available in a channel that it was not assigned to. Document prefix: " + user_1_doc_prefix + ". The document: " + doc)
+        if shared_doc_prefix in doc:
+            shared_found_user_2 = True
+        if doc not in wildcard_user_docs_ids:
+            pytest.fail("The document " + doc + " was not accessiable even though the user was given all documents access")
```
hthomps-cb, line 227: typo: accessiable -> accessible

gilkal:
> 

Thanks, fixed.

testsuites/syncgateway/functional/tests/test_cbs_collections.py
gilkal: Thanks, fixed.

testsuites/syncgateway/functional/tests/test_cbs_collections.py
gilkal: Thanks, fixed

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -109,11 +112,138 @@ def test_document_only_under_named_scope(scopes_collections_tests_fixture, teard
...
+
+    # 5. Check that the users see the shared document in their channels
+    assert (shared_found_user_1 and shared_found_user_2), "The shared document was not found for one of the users. user1: " + shared_found_user_1 + " user2: " + shared_found_user_2
+    assert (shared_doc_prefix in doc for doc in wildcard_user_docs_ids), "The shared document was not accessiable VIA the wildcard channel"
```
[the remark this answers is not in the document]
gilkal: Weirdly, that assert also always return true, no matter what I gave it (assert (shared_doc_prefix in doc for doc in wildcard_user_docs_ids)!

I'm not quite sure what was going on there, but I think now it's working.

Thanks for picking that up!

hthomps-cb (approved):

PR #1981, opened 2022-11-17
Fix/test document only under named scope

muntazim-aliCB:
- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

Fix test for reading document from collection without specifying scope.


gilkal (approved):
Thank you for picking it up and fixing it!

PR #1985, opened 2022-11-28
Test/restricted_collection

muntazim-aliCB:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

Sync Gateway scopes and collections test for verifying documents in a server only collection are not pulled by Sync Gateway.

### DIFF
--- a/keywords/couchbaseserver.py
+++ b/keywords/couchbaseserver.py
@@ -6,7 +6,7 @@ from datetime import timedelta
 from requests.exceptions import ConnectionError, HTTPError, ChunkedEncodingError
 from requests import Session
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
-from couchbase.exceptions import CouchbaseException, DocumentNotFoundException
+from couchbase.exceptions import CouchbaseException, DocumentNotFoundException, DocumentExistsException
 from couchbase.cluster import QueryIndexManager, PasswordAuthenticator, ClusterTimeoutOptions, ClusterOptions, Cluster
 import keywords.constants
 from keywords.remoteexecutor import RemoteExecutor
@@ -16,6 +16,7 @@ from keywords.utils import log_r, log_info, log_debug, log_error, hostname_for_u
 from keywords.utils import version_and_build, random_string
 from keywords import types
 from utilities.cluster_config_utils import is_x509_auth, get_cbs_version, is_magma_enabled, is_cbs_ce_enabled, get_cluster
+from libraries.data import doc_generators
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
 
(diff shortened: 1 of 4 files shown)


keywords/couchbaseserver.py
```diff
@@ -1182,16 +1183,47 @@ def rebalance_server(self, cluster_servers):
...
-        connection_url = "couchbase://{}?ipv6=allow".format(cbs_ip)
+        doc_body = doc_generators.simple()
+        doc_body["id"] = doc_id
+        try:
+            result = collection_client.insert(doc_id, doc_body)
+        except DocumentExistsException as e:
+            raise Exception("Tried to insert document that already exists: " + str(e)) from e
+        return result
+
+    def get_document(self, bucket, scope, collection, doc_id, ipv6=False):
```
gilkal, line 1200: You can consider adding default values to the scope and collection (like the "_default" scopes/collection). It's not critical, maybe even just a personal preference.

keywords/couchbaseserver.py
```diff
@@ -1182,16 +1183,47 @@ def rebalance_server(self, cluster_servers):
...
+        collection_client = sdk_client.scope(scope).collection(collection)
+
+        try:
+            result = collection_client.get(doc_id)
+        except DocumentNotFoundException as e:
+            raise Exception("Tried to fetch document that does not exist: " + str(e)) from e
+        return result.content_as[dict]
+
+
+def choose_connection_url(ssl_enabled, ipv6, host):
```
gilkal, line 1213: I feel like I've seen similar code to this somewhere, but can't find it. Was this copied from somewhere? Can we make some sort of sharing if that's the case?

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -261,6 +261,48 @@ def test_collection_channels(scopes_collections_tests_fixture):
...
+    cb_server.create_collection(bucket, scope, second_collection)
+
+    doc_1_key = "doc_1" + random_suffix
+    doc_2_key = "doc_2" + random_suffix
+
+    # 2. Add a document to each collection
+    cb_server.add_simple_document(bucket, scope, collection, doc_1_key)
+    cb_server.add_simple_document(bucket, scope, second_collection, doc_2_key)
+
+    assert(cb_server.get_document(bucket, scope, collection, doc_1_key)["id"] == doc_1_key)
```
gilkal, line 291: I would consider removing these 2 checks. You're right that it's a good practice to check that the setup was correct, but if that functionality doesn't work then there's no point to run the test anyway, and it'll be just one failure out of mass failures.

If you do decide to keep them, you can add more value by giving a verbal description, for. example: "There was a problem during the test setup: the document was not added to the server"

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -261,6 +261,48 @@ def test_collection_channels(scopes_collections_tests_fixture):
...
+
+    # 3. Sync one collection to SGW
+    db_config = {"bucket": bucket, "scopes": {scope: {"collections": {collection: {}}}}, "num_index_replicas": 0,
+                 "import_docs": True, "enable_shared_bucket_access": True}
+    admin_client.post_db_config(db, db_config)
+    admin_client.wait_for_db_online(db, 60)
+
+    # 4. Check that documents in the server restricted collection are not accesible via SGW
+    all_docs = sg_client.get_all_docs(sg_admin_url, db)
+    assert(all_docs["total_rows"] == 1)
```
gilkal, line 302: It'll be good to give a description in case the test fails, so it'll be easier from just looking at the logs to understand what's going on. For example: "The number of documents is different from the expected. This could be because restricted files are being picked up when they shouldn't" (you probably have a more concise and informative description than this).

utilities/cluster_config_utils.py
```diff
@@ -38,12 +38,14 @@ def write(self, fp):
...
 
 
-def get_cluster(url, bucket_name):
+def get_cluster(url, bucket_name, get_bucket=False):
     timeout_options = ClusterTimeoutOptions(kv_timeout=timedelta(seconds=180), query_timeout=timedelta(seconds=300),
                                             config_total_timeout=timedelta(seconds=600))
     options = ClusterOptions(PasswordAuthenticator("Administrator", "password"), timeout_options=timeout_options)
     cluster = Cluster(url, options)
     cluster = cluster.bucket(bucket_name)
+    if get_bucket:
```
gilkal, line 47: Just curious: what is that for?

keywords/couchbaseserver.py
muntazim-aliCB: I can't find anything similar in the testkit either.

utilities/cluster_config_utils.py
muntazim-aliCB: wanted to access the cluster at the bucket level rather than the default collection, but I think I've found another way around it so this can be removed

keywords/couchbaseserver.py
muntazim-aliCB: Added

testsuites/syncgateway/functional/tests/test_cbs_collections.py
muntazim-aliCB: Decided to keep them and add error messages

testsuites/syncgateway/functional/tests/test_cbs_collections.py
muntazim-aliCB: some descriptions for test failure added, let me know if they need more detail

gilkal (approved):

PR #2003, opened 2022-12-14
Update requirements.txt

saigon:
We will not install couchbase==3.2.7 in requirement.  In stead, we install it in setup.sh because pip version 20.1.1 will install couchbase 3.2.7 ok on python 3.7.x Downgrade flake8 to 3.3.0 because 3.9.x break android api tests

#### Fixes #.

- [ ] Ran `flake8`
- [ ] Ran `run_repo_tests.sh`

- 
-

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ docker-pycreds==0.2.1
 docutils==0.12
 ecdsa==0.13
 enum34==1.1.6
-flake8==3.3.0
+flake8==3.9.2
 futures==3.1.1
 idna==2.5
 ipaddress==1.0.18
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -10,14 +10,14 @@ botocore==1.4.5
...
-couchbase==3.2.7
+#couchbase==3.2.7
 cryptography==37.0.4
 docker==2.2.1
 docker-pycreds==0.2.1
 docutils==0.12
 ecdsa==0.13
 enum34==1.1.6
-flake8==3.9.2
+flake8==3.3.0
```
abhayayay, line 20: I hope this doesn't break ios, c, and sync-gateway jobs. 
I will check it in my system. Approving as of now, will let you know if any error arises

abhayayay (approved):

PR #2012, opened 2022-12-28
Update conftest.py

abhayayay:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Removed conflicting variable from P2P conftest.py


gilkal (approved):
Looking good!

Next time, could add an explanation and check the flake8/run_repo_tests? It's just easier that way to understand a change, in general.

Thanks,

Gilad

PR #2027, opened 2023-01-13
Add task to fetch trace logs

muntazim-aliCB:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

### DIFF
--- a/libraries/provision/ansible/playbooks/tasks/fetch-sync-gateway-logs.yml
+++ b/libraries/provision/ansible/playbooks/tasks/fetch-sync-gateway-logs.yml
@@ -70,3 +70,15 @@
   # Do not validate checksum as this is just a passive grab of the logs. Since they are continuously being appended to, the checksum may fail
   fetch: src=/home/sync_gateway/logs/sg_warn.log dest=/tmp/sg_logs/{{ inventory_hostname }}/ fail_on_missing=yes flat=yes validate_checksum=no
   when: sw.stat.exists
+
+- name: SYNC GATEWAY | Check if sg_trace.log is present
+  become: yes
+  stat:
+    path: '/home/sync_gateway/logs/sg_trace.log'
+  register: st
+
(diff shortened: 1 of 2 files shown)


libraries/provision/ansible/playbooks/tasks/fetch-sync-gateway-logs.yml
```diff
@@ -70,3 +70,15 @@
...
+- name: SYNC GATEWAY | Check if sg_trace.log is present
+  become: yes
+  stat:
+    path: '/home/sync_gateway/logs/sg_trace.log'
+  register: st
+
+- name: SYNC GATEWAY | Fetch sg_trace.log
+  become: yes
+  # Do not validate checksum as this is just a passive grab of the logs. Since they are continuously being appended to, the checksum may fail
+  fetch: src=/home/sync_gateway/logs/sg_trace.log dest=/tmp/sg_logs/{{ inventory_hostname }}/ fail_on_missing=yes flat=yes validate_checksum=no
```
gilkal, line 83: Are we sure we want to fail if the trace logs are missing? We have quite a few provisioning problems, could that potentially add one more?

muntazim-aliCB: It should be fine, as we are checking the existence of the logs above. Should they not exist then the fetch is not attempted, the fail_on_missing parameter here is checking the ability of the runner to actually read the file.

By default we have the trace logs disabled, so the fetch is only being run when we explicitly set the trace-logs cmdline arg, in which case we are running with the intention of explicitly retrieving the trace logs. It makes sense then to fail if we cannot retrieve them.

gilkal: Yep, that's a good point, thanks!

gilkal (approved):

PR #2066, opened 2023-04-24
CM-1109: Add polling to an async assertions section

gilkal:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- Add retries to the critical sections that fail in the tests, due to timing issues.
- It's not the ideal approach to fix these. 

    * The best thing would a generic SGW query that can tell when the sync is done. I don't think there's something like that.
    * I could try to create a function the does the retires and accepts a function handle. however, that would require sending too many arguments, which in itself "ugly" so I'd rather not do that.
    * I couldn't find any easy generic Python way to do that. There are 3rd party dependencies, but I would rather not add unnecessary dependencies that could cause future incompatibility.

### DIFF
--- a/testsuites/syncgateway/functional/tests/test_cbs_collections.py
+++ b/testsuites/syncgateway/functional/tests/test_cbs_collections.py
@@ -12,7 +12,7 @@ from requests.auth import HTTPBasicAuth
 from requests.exceptions import HTTPError
 from keywords import document
 from libraries.data import doc_generators
-from time import sleep
+from time import sleep, time
 
 # test file shared variables
 bucket = "data-bucket"
@@ -329,9 +329,9 @@ def test_restricted_collection(scopes_collections_tests_fixture):
     admin_client.wait_for_db_online(db, 60)
 
     # 4. Check that documents in the server restricted collection are not accesible via SGW
-    retries = 0
-    max_retries = 15
-    while retries < max_retries:
+    timeout = 15  # 15 seconds
+    start = time.time()
+    while True:
         all_docs_ids = []
         for row in (sg_client.get_all_docs(sg_admin_url, db, scope=scope, collection=collection)["rows"]):
             all_docs_ids.append(row["id"])
@@ -342,10 +342,9 @@ def test_restricted_collection(scopes_collections_tests_fixture):
             assert(doc_3_key not in all_docs_ids), "Sync Gateway contains document from server restricted collection. Document ID " + doc_3_key + "under " + bucket + "." + scope + "." + third_collection
             break
         except Exception as e:
-            if retries == max_retries - 1:
+            if time.time() - start > timeout:
                 raise e
         sleep(1)
-        retries = retries + 1
 
 
 @pytest.mark.syncgateway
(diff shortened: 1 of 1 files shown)


testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -328,15 +329,23 @@ def test_restricted_collection(scopes_collections_tests_fixture):
...
-
-    for row in (sg_client.get_all_docs(sg_admin_url, db, scope=scope, collection=collection)["rows"]):
-        all_docs_ids.append(row["id"])
-    for row in (sg_client.get_all_docs(sg_admin_url, db, scope=scope, collection=second_collection)["rows"]):
-        all_docs_ids.append(row["id"])
-
-    assert(len(all_docs_ids) == 2), "Number of expected documents in Sync Gateway database does not match expected. Expected 2; Found " + str(len(all_docs_ids))
-    assert(doc_3_key not in all_docs_ids), "Sync Gateway contains document from server restricted collection. Document ID " + doc_3_key + "under " + bucket + "." + scope + "." + third_collection
+    retries = 0
+    max_retries = 15
```
muntazim-aliCB, line 333: Could we maybe make this based on a max time rather than number of calls? That way we can say that the documents weren't synced after X seconds, which seems more useful than an arbitrary number of retries.

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -619,8 +628,18 @@ def test_collection_stats(scopes_collections_tests_fixture):
     assert("403" in str(e)), "User without access to collection should generate 403 HTTPError when trying to GET document. Instead got: \n" + str(e)
 
     # 5. Verify stats reflect changes from API calls correctly
-    new_stats = sg_client.get_expvars(sg_admin_url, admin_auth)
-    verify_collection_stats(db, scope, second_collection, third_collection, new_stats)
+    retries = 0
+    max_retries = 15
```
muntazim-aliCB, line 632: Same comment here

testsuites/syncgateway/functional/tests/test_cbs_collections.py
```diff
@@ -328,15 +329,23 @@ def test_restricted_collection(scopes_collections_tests_fixture):
...
+        for row in (sg_client.get_all_docs(sg_admin_url, db, scope=scope, collection=second_collection)["rows"]):
+            all_docs_ids.append(row["id"])
+        try:
+            assert(len(all_docs_ids) == 2), "Number of expected documents in Sync Gateway database does not match expected. Expected 2; Found " + str(len(all_docs_ids))
+            assert(doc_3_key not in all_docs_ids), "Sync Gateway contains document from server restricted collection. Document ID " + doc_3_key + "under " + bucket + "." + scope + "." + third_collection
+            break
+        except Exception as e:
+            if retries == max_retries - 1:
+                raise e
+        sleep(1)
```
muntazim-aliCB, line 347: If using a time based approach may make sense to have an exponential backoff rather than making an API request every other second

gilkal: Thanks, I agree using time is better than retry, fixed! About reporting the timeout, it slightly complicates things and the error message is informative, so how about leaving that bit out?

testsuites/syncgateway/functional/tests/test_cbs_collections.py
gilkal: Thanks, I agree using time is better than retry, fixed! About reporting the timeout, it slightly complicates things and the error message is informative, so how about leaving that bit out?

muntazim-aliCB (approved):

PR #2091, opened 2023-10-03
Handle new supportlib distribution method

bmeike:
I believe this change will correct the failure caused by the new SupportLib distribution mechanism for Java.


gilkal (approved):
I tested it in Jenkins - it looks OK. I also ran flake8

PR #2112, opened 2024-03-04
Add/cbl platform flags

gilkal:
#### Fixes #.

- [x] Ran `flake8`
- [x] Ran `run_repo_tests.sh`

- More framework adjustments to support running SGW/CBL on Debian12.
- Tested on iOS and Android sanity. More changes may be required.

### DIFF
--- a/keywords/SyncGateway.py
+++ b/keywords/SyncGateway.py
@@ -684,18 +684,22 @@ class SyncGateway(object):
         shut down the sync gateway at that url
         """
         ansible_runner = AnsibleRunner(cluster_config)
-
+        extra_vars = {}
+        if "debian" in self.sg_platform.lower():
+            extra_vars["ansible_distribution"] = self.sg_platform.capitalize()
         if url is not None:
             target = hostname_for_url(cluster_config, url)
             log_info("Shutting down sync_gateway on {} ...".format(target))
             status = ansible_runner.run_ansible_playbook(
                 "stop-sync-gateway.yml",
+                extra_vars=extra_vars,
                 subset=target
             )
         else:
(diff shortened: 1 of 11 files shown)


keywords/SyncGateway.py
```diff
@@ -683,18 +681,20 @@ def stop_sync_gateways(self, cluster_config, url=None):
         shut down the sync gateway at that url
         """
         ansible_runner = AnsibleRunner(cluster_config)
-
+        extra_vars = {}
```
muntazim-aliCB, line 684: Is this being used? I see that the default argument for `ansible_runner.run_ansible_playbook()` is already an empty  dict, and it doesn't seem like this variable is ever changed?

libraries/provision/ansible/playbooks/tasks/create-sync-gateway-user.yml
```diff
@@ -1,8 +1,3 @@
-# Add sync_gateway user on Centos/linux/ubuntu/debian
```
muntazim-aliCB, line 1: This passes for the sanity tests I assume, but are there other tests that need a SG user that no longer pass without this?

gilkal: The sync_gateway user is created by default in the installation, so I think we don't need it. The problem is that the old Ansible version is not compatible with Debian12, causing this user creation to fail.

keywords/SyncGateway.py
gilkal: Thanks, fixed!

muntazim-aliCB (approved):

